PR #2, opened 2018-05-30
codereview in form of PR

pongad:


server/showcase.go
```diff
@@ -15,40 +15,47 @@
...
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
+type uniqID struct {
+	i int64
+}
+
+func (u *uniqID) id() int64 {
+	return atomic.AddInt64(&u.i, 1) - 1
```
pongad: Using current time as ID can lead to weird failures during clock resets and concurrent calls. Unlikely to cause problem since we're using it to test, but thought we might as well harden against it.

server/showcase.go
```diff
@@ -15,40 +15,47 @@
...
 		operationStore: opStore,
-		nowF:           time.Now,
 		sleepF:         time.Sleep,
 	}
 }
 
 type showcaseServerImpl struct {
+	uid uniqID
+
+	mu             sync.Mutex
```
pongad: The mutex is idiomatically at the top of the "block" containing things it's protecting.

landrito:
Thanks @pongad!

PR #4, opened 2018-06-28
Trivial change to baseline a PR.

lukesneeringer:
This is a one-line change to open up the file for PR comments. :-p

### DIFF
--- a/showcase.proto
+++ b/showcase.proto
@@ -22,6 +22,7 @@ import "google/rpc/status.proto";
 
 package google.showcase.v1;
 
+
 // A service to showcase Generated API Client features and common API patterns
 // used by Google.
 service Showcase {
--- a/showcase.proto
+++ b/showcase.proto
@@ -28,38 +28,38 @@ package google.showcase.v1;
 service Showcase {
   // A method used to show unary methods. This method will return the message
   // that was given.
-  rpc Echo(EchoRequest) returns (EchoResponse) {}
+  rpc Echo(EchoRequest) returns (EchoResponse);
 
   // A method used to show server-side streaming methods. This method will
   // split the given content into words and will pass each word back through
   // the stream.
-  rpc Expand(ExpandRequest) returns (stream EchoResponse) {}
+  rpc Expand(ExpandRequest) returns (stream EchoResponse);
 
   // A method used to show client-side streaming methods. This method will
   // collect the contents given to it. When the stream is closed by the client,
   // this method will return the a concatenation of the strings passed to it.
-  rpc Collect(stream EchoRequest) returns (EchoResponse) {}
+  rpc Collect(stream EchoRequest) returns (EchoResponse);
 
   // A method that will be used to show bidirectional streaming methods. Upon
   // receiving a request on the stream, the same content will be passed back
   // on the stream.
 
-  rpc Chat(stream EchoRequest) returns (stream EchoResponse) {}
+  rpc Chat(stream EchoRequest) returns (stream EchoResponse);
 
   // This method is to show how a client handles a request timing out. The
   // server will wait the requested amount of response_delay and then return
   // the response specified in the request.
-  rpc Timeout(TimeoutRequest) returns (TimeoutResponse) {}
+  rpc Timeout(TimeoutRequest) returns (TimeoutResponse);
 
   // This method is used to setup the Retry method. The given list of retry behavior
   // will be bound to an ID defined by the server. Subsequent Retry requests
   // of the given ID will respond with the behavior specified in this request.
-  rpc SetupRetry(SetupRetryRequest) returns (RetryId) {}
+  rpc SetupRetry(SetupRetryRequest) returns (RetryId);
 
   // This method is used to show how a client handles errors that can be
   // retried. Requests to a given RetryId will respond in the way specified
   // in SetupRetry method.
-  rpc Retry(RetryId) returns (google.protobuf.Empty) {}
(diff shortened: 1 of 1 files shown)


showcase.proto
```diff
@@ -22,6 +22,7 @@ import "google/rpc/status.proto";
 
 package google.showcase.v1;
```
lukesneeringer: This should be `v1alpha1`. :-)
Also, perhaps `google.api.showcase`?

showcase.proto
```diff
@@ -22,43 +22,44 @@ import "google/rpc/status.proto";
...
 
+
 // A service to showcase Generated API Client features and common API patterns
 // used by Google.
 service Showcase {
   // A method used to show unary methods. This method will return the message
   // that was given.
-  rpc Echo(EchoRequest) returns (EchoResponse) {}
+  rpc Echo(EchoRequest) returns (EchoResponse);
```
lukesneeringer: Sanity check: I assume you intentionally skipped having `google.api.http` annotations since you do not (at least not yet anyway) have an HTTP server equivalent?

showcase.proto
```diff
@@ -22,43 +22,44 @@ import "google/rpc/status.proto";
...
 
   // A method used to show server-side streaming methods. This method will
   // split the given content into words and will pass each word back through
   // the stream.
-  rpc Expand(ExpandRequest) returns (stream EchoResponse) {}
+  rpc Expand(ExpandRequest) returns (stream EchoResponse);
```
lukesneeringer: Despite the duplication, can there be a separate `ExpandResponse`? I understand why you did this, but our general guidance is a unique request/response for most calls, and since this may be many people's first exposure, I think this detail matters.

showcase.proto
```diff
@@ -22,43 +22,44 @@ import "google/rpc/status.proto";
...
   // server will wait the requested amount of response_delay and then return
   // the response specified in the request.
-  rpc Timeout(TimeoutRequest) returns (TimeoutResponse) {}
+  rpc Timeout(TimeoutRequest) returns (TimeoutResponse);
 
   // This method is used to setup the Retry method. The given list of retry behavior
   // will be bound to an ID defined by the server. Subsequent Retry requests
   // of the given ID will respond with the behavior specified in this request.
-  rpc SetupRetry(SetupRetryRequest) returns (RetryId) {}
+  rpc SetupRetry(SetupRetryRequest) returns (RetryId);
```
lukesneeringer: Suggestion:

```protobuf
rpc CreateRetryDemonstration(CreateRetryDemonstrationRequest) returns (RetryDemonstration);
```

That allows it to expand to updating behavior in the future if necessary. (I do not see it on the horizon, but I can see how it is possible.) It also means it can follow the standard method convention.

This would also mean the next method needs to be a verb-noun (which is appropriate anyway); something like `AttemptRetryDemonstration` or `PokeRetryDemonstration`.

showcase.proto
```diff
@@ -69,26 +70,27 @@ service Showcase {
...
   // is complete, else the server will return a completed operation that will
   // either denote that the operation has succeeded or failed.
-  rpc Longrunning(LongrunningRequest) returns (google.longrunning.Operation) {}
+  rpc Longrunning(LongrunningRequest) returns (google.longrunning.Operation);
 
   // This method is used to show how a client will handle a method that lists
   // responses in a paginated method. The request will specify an ID and the
   // maximum number of responses to return. If the ID has not been seen before
   // the server will mark the ID specified and the maximum number to be
   // returned. The page token will be a stringified integer denoting the
```
lukesneeringer: Page tokens should be opaque. Is there a reason to not do that here?

showcase.proto
```diff
@@ -69,26 +70,27 @@ service Showcase {
...
   // the number to start the page at.
-  rpc Pagination(PaginationRequest) returns (PaginationResponse) {}
+  rpc Pagination(PaginationRequest) returns (PaginationResponse);
```
lukesneeringer: This should be a List method of some kind, yes?

showcase.proto
```diff
@@ -69,26 +70,27 @@ service Showcase {
...
 
   // This method is used to show how a client will handle configuration to
   // flatten a request. The server will simply echo the request and
```
lukesneeringer: and...?

showcase.proto
```diff
@@ -69,26 +70,27 @@ service Showcase {
...
-  rpc ParameterFlattening(ParameterFlatteningMessage) returns (ParameterFlatteningMessage) {}
+  rpc ParameterFlattening(ParameterFlatteningMessage) returns (ParameterFlatteningMessage);
```
lukesneeringer: I am struggling to understand what this RPC does or why it exists in this form.

showcase.proto
```diff
@@ -69,26 +70,27 @@ service Showcase {
...
 
   // This method is used to show how a client will handle request strings that
   // are configured to follow a certain pattern or set of patterns. The server
   // will simply echo the request.
-  rpc ResourceName(ResourceNameMessage) returns (ResourceNameMessage) {}
+  rpc ResourceName(ResourceNameMessage) returns (ResourceNameMessage);
```
lukesneeringer: I am struggling to understand what this RPC does or why it exists in this form.

showcase.proto
lukesneeringer: (This applies throughout these.)

landrito:
This has gotten moldy closing due to recent design reviews.

PR #6, opened 2018-08-24
Server-side validation MVP

landrito:
This is the minimum viable product for server side validation.


landrito:
Disregard the changes to remove pagesize override which will be added in #9 and the changes to add the compile protos script which will be added in #8. I will rebase after those are merged.

server/session.go
```diff
@@ -0,0 +1,273 @@
...
+	DeleteTest(name string) error
+	AddAnswers(name string, answers []string) error
+	TestAnswers(name string, answers []string) error
+}
+
+var instance Session
+var once sync.Once
+
+func GetSessionSingleton() Session {
+	once.Do(func() {
```
pongad: No need to use `once.Do`. Just
```
var Instance Session = &sessionImpl{...}
```
at top level

server/session.go
```diff
@@ -0,0 +1,273 @@
...
+			tests: map[string]Test{},
+		}
+	})
+	return instance
+}
+
+type sessionImpl struct {
+	name string
+
+	mu    *sync.Mutex
```
pongad: mutex doesn't have to be a pointer.

server/session.go
```diff
@@ -0,0 +1,273 @@
...
+
+	numOptional := 0
+	numOptionalSkipped := 0
+	numOptionalFailed := 0
+
+	errors := []*pb.ReportSessionResponse_Issue{}
+	warnings := []*pb.ReportSessionResponse_Issue{}
+
+	for _, test := range s.tests {
+		expectationLevel := test.GetExpectationLevel()
```
pongad: I think `expLvl` is a better name, but up to you :P

server/session.go
```diff
@@ -0,0 +1,273 @@
...
+	mu    *sync.Mutex
+	tests map[string]Test
+}
+
+func (s *sessionImpl) GetName() string {
+	return s.name
+}
+
+func (s *sessionImpl) GetReport() *pb.ReportSessionResponse {
+	numRequired := 0
```
pongad: I'd wrap it like this:
```
var (
numRequired int
numRequiredSkipped int
//...
errors []*pb.ResportSessionResponse_Issue // no need to init array, append already knows what to do
)
```

Also consider creating a type `type result struct {test, skipped, failed int}`

server/session.go
```diff
@@ -0,0 +1,273 @@
...
+					numOptionalSkipped = numOptionalSkipped + 1
+				} else {
+					numOptionalFailed = numOptionalFailed + 1
+				}
+			}
+		}
+	}
+
+	var state pb.ReportSessionResponse_State
+	if numRequired == 0 || (numRequiredFailed == 0 && numRequiredSkipped == 0) {
```
pongad: shouldn't `numRequiredFailed == 0 && numRequiredSkipped ==0` be sufficient?

server/session.go
```diff
@@ -0,0 +1,273 @@
...
+	numOptionalFailed := 0
+
+	errors := []*pb.ReportSessionResponse_Issue{}
+	warnings := []*pb.ReportSessionResponse_Issue{}
+
+	for _, test := range s.tests {
+		expectationLevel := test.GetExpectationLevel()
+		issue := test.GetIssue()
+
+		if expectationLevel == pb.Test_REQUIRED {
```
pongad: If you create the `result` type, you can do
```
var r *result
switch expectationLevel {
case pb.Test_REQUIRED:
  r = &resultRequired
// ...
}

r.test++
if issue != nil {
  r.issues = append(r.issues, issue)
  if issue.Type == SKIPPED {
    r.skipped++
  } else {
    r.failed++
  }
}
```

Using pointer is also OK without the result type, but it's more awkward.

server/testing_service_test.go
```diff
@@ -0,0 +1,163 @@
+// Copyright 2018 Google LLC
```
pongad: I won't stop you but I don't think these tests are actually necessary. `testingServerImpl` essentially forwards things to session.

server/testing_service.go
```diff
@@ -0,0 +1,57 @@
...
+	err := s.session.DeleteTest(in.Name)
+	if err != nil {
+		return nil, err
+	}
+	return &empty.Empty{}, nil
+}
+
+func (s *testingServerImpl) RegisterTest(ctx context.Context, in *pb.RegisterTestRequest) (*empty.Empty, error) {
+	name := in.Name
+	name = strings.TrimPrefix(name, "/sessions/-")
```
pongad: Do we want to error if the name doesn't start with the prefix?

landrito: Eventually. I think what is present is fine for now until I have the multiple sessions story. For now we can simply defer to the session to do the error handling.

I'll make a comment in code.

server/testing_service_test.go
landrito: Agreed. Admittedly, I just wanted to keep 100% test coverage.

server/session.go
landrito: Ah yes that's correct. Will change.

server/session.go
landrito: Ah this is much better thanks! Done.

server/session.go
landrito: Done.

server/session.go
landrito: Done.

server/session.go
landrito: Done!

server/session.go
landrito: Done.

server/session.go
landrito: Done.

PR #8, opened 2018-08-28
Add proto compilation script

landrito:
@pongad can you take a look?


cmd/compile_protos/main.go
```diff
@@ -0,0 +1,81 @@
...
+import (
+	"log"
+	"os"
+	"os/exec"
+	"path/filepath"
+)
+
+func main() {
+	// Check if protoc is installed.
+	cmd := exec.Command("protoc", "--version")
```
pongad: `exec.Command(...).Run()` is perfectly OK to chain.

cmd/compile_protos/main.go
```diff
@@ -0,0 +1,81 @@
...
+	if err := cmd.Run(); err != nil {
+		log.Fatal("Error: 'protoc' is expected to be installed on the path.")
+	}
+
+	// Get proto dependencies
+	gopath := os.Getenv("GOPATH")
+	showcaseDir := gopath + "/" + "src/github.com/googleapis/gapic-showcase"
```
pongad: Here and below: Prefer `filepath.Join` here in case we run on Windows?

cmd/compile_protos/main.go
landrito: Done.

cmd/compile_protos/main.go
```diff
@@ -58,8 +57,7 @@ func main() {
 		"google",
 		"showcase",
 		"v1alpha2")
-	err = exec.Command("mkdir", "-p", protoDest).Run()
-	if err != nil {
+	if err = exec.Command("mkdir", "-p", protoDest).Run(); err != nil {
```
pongad: Up to you, but you can use `os.MkdirAll` here.

landrito:
Thanks for the review @pongad!

Issue #10, opened 2018-08-30
Add logging for additional RPC types


jbolinger:
Unary requests are logged to stdout but other types are not (streaming, etc.): 

https://github.com/googleapis/gapic-showcase/blob/c7d0008906adbbad38af53ce8a62d71a2e85b6f5/cmd/server/main.go#L91

Similar logging for additional RPC types would be helpful.

landrito:
Resolved in #11

jbolinger:
Thanks! I'll try to update soon.

PR #11, opened 2018-09-08
Refactor CLI code

landrito:
This PR refactors the CLI code. This refactor is for a couple of different reasons:

1. Add a bunch of usage messages to the command line interface.
2. Make it such that adding commands to the command line interface is much easier to do because the usage message handling is handled by a cli manager.


cmd/daemon.go
```diff
@@ -0,0 +1,117 @@
...
+		status, err := d.Install("start", "--port", InstallPort)
+		if err != nil {
+			ErrLog.Println(status, "\nError: ", err)
+			os.Exit(1)
+		}
+		StdLog.Println(status)
+	},
+}
+
+var daemonStartCmd = &cobra.Command{
```
pongad: I'm not very familiar with the daemon. What's the difference between installing and starting?

landrito: For linux:

`Install` creates a shell script that is placed at [/etc/init.d/](https://askubuntu.com/questions/5039/what-is-the-difference-between-etc-init-and-etc-init-d) I assume this is the common place to put daemon-like executables. 

`Start` calls the shell script to invoke the go program.

For other platforms, the commands do behaviorally similar things but with what is idiomatic for the platform.

landrito:
This change also adds a StreamInterceptor which should resolve #10.

cmd/start.go
```diff
@@ -0,0 +1,143 @@
...
+	if err == nil {
+		stdLog.Printf("    Returning Response: %+v\n", resp)
+	} else {
+		stdLog.Printf("    Returning Error: %+v\n", err)
+	}
+	stdLog.Println("")
+	return resp, err
+}
+
+type loggingServerStream struct {
```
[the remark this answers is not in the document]
landrito: Ah found the way.

cmd/daemon.go
pongad: Ah

PR #15, opened 2018-09-14
fix flattening config

jbolinger:
The 0.5 release removed a field from the proto, but left it in the `method_signature` annotation.

### DIFF
--- a/schema/echo.proto
+++ b/schema/echo.proto
@@ -91,7 +91,7 @@ service Echo {
       body: "*"
     };
     option (google.api.method_signature) = {
-      fields: ["max_response", "page_size_override"]
+      fields: ["max_response"]
     };
   }
 }


jbolinger:
I think you could use the Kotlin generator (or another) if you wanted to catch things like this. It will try and compile the generated code and that would fail in cases like this.

For example, try running this from the showcase project directory:

```bash
$ mkdir my-output
$ docker run --rm \
         --mount type=bind,source="$(pwd)"/schema,target=/proto \
         --mount type=bind,source="$(pwd)"/my-output,target=/generated \
    gcr.io/kotlin-gapic/kgen
```

So, you could put something like that in your CI workflow to catch these (I'd probably pin the image to a known tag though and not use `latest`, at least for now).

PR #16, opened 2018-09-17
Add Kotlin Smoke Test

landrito:


landrito:
Hey @jbolinger any idea how I could fix the errors I'm getting in [circleci](https://circleci.com/gh/googleapis/gapic-showcase/105?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)?

jbolinger:
I suspect that it's because the the testing proto doesn't have `option java_multiple_files = true;`

https://github.com/googleapis/gapic-showcase/blob/4e75c256f9a0c8e1a62ab210125c885204e5bbe0/schema/testing.proto#L23

I haven't spent any time testing the generator without that option, so it's likely producing invalid code.

.circleci/config.yml
```diff
@@ -54,6 +56,28 @@ jobs:
           command: bash <(curl -s https://codecov.io/bash)
           when: always
 
+  kotlin-smoke-test:
+    machine: true
+    steps:
+      - checkout
+      - run:
+          name: Get kotlin generator
+          command: docker pull gcr.io/kotlin-gapic/kgen
```
landrito: @jbolinger I am having a hard time finding the appropriate tag to use for this docker image. Which should I use?

landrito:
Thanks @jbolinger, that worked! This is ready for review when you have the chance.

.circleci/config.yml
jbolinger: Are you able to access this page? https://console.cloud.google.com/gcr/images/kotlin-gapic/GLOBAL/kgen

You can see the tags there (I suspect that we'll move things to docker hub as things mature). You can pin to a recent tag or try to stay up to date with latest. I try not to break things but it's early days :)

PR #19, opened 2018-09-20
Update Documentation

landrito:
The old READMEs were a bit outdated.


README.md
```diff
@@ -1,131 +1,236 @@
...
+$ go install github.com/googleapis/gapic-showcase
+$ gapic-showcase start
+> 2018/09/19 02:13:09 Showcase listening on port: :7469
+```
+
+<!---
+TODO(landrito): figure out a blessed way to install by version using go
+commands.
+-->
+_* Bare in mind this is not a versioned installation so no versioning guarantees
```
timburks: "Bear" :)

landrito: Fixed!

jbolinger:
@landrito it should be fixed, you just need to update to a newer image. I'll send you a PR to update and verify. 

Before I do that let me pin the version of kgax in the generator so it's not using SNAPSHOTs and you get repeatable checks. I was just being lazy since I'm not yet doing any real releases yet...

Issue #25, opened 2018-09-27
Echo service specifies a port number that isn't used.


andreamlin:
service Echo specifies `google.api.default_host` to be `localhost:7469`, but this value isn't used in the generated clients. The actual value used is `localhost:443`, where `localhost` is pulled from the service config and [`443` is the service port given by gapic-generator](https://github.com/googleapis/gapic-generator/blob/5480c68d4b1020447f3926703fa184e745372ee3/src/main/java/com/google/api/codegen/config/ProtoApiModel.java#L56)

In order to transition away from service config and towards using proto annotations, the proto annotation value for `google.api.default_host` must match the [`name`](https://github.com/googleapis/gapic-generator/blob/5480c68d4b1020447f3926703fa184e745372ee3/src/main/java/com/google/api/codegen/config/ProtoApiModel.java#L50) field of the service config, which for showcase.yaml is currently [`localhost`](https://github.com/googleapis/gapic-generator/blob/5480c68d4b1020447f3926703fa184e745372ee3/src/test/java/com/google/api/codegen/testsrc/showcase.yaml#L3).

andreamlin:
Turns out that we should be using this port, but gapic-generator hadn't been properly been parsing out the port. Fixed in https://github.com/googleapis/gapic-generator/pull/2350

PR #26, opened 2018-09-27
Change echo service default_host to localhost, without port

andreamlin:
Resolves https://github.com/googleapis/gapic-showcase/issues/25

See https://github.com/googleapis/gapic-generator/pull/2346 for gapic-generator CI build using this PR.

### DIFF
--- a/schema/echo.proto
+++ b/schema/echo.proto
@@ -36,7 +36,7 @@ option (google.api.metadata) = {
 service Echo {
   // This service is meant to only run locally on the port 7469 (keypad digits
   // for "show").
-  option (google.api.default_host) = "localhost:7469";
+  option (google.api.default_host) = "localhost";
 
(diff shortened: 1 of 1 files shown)


schema/echo.proto
```diff
@@ -36,7 +36,7 @@ option (google.api.metadata) = {
 service Echo {
   // This service is meant to only run locally on the port 7469 (keypad digits
   // for "show").
-  option (google.api.default_host) = "localhost:7469";
+  option (google.api.default_host) = "localhost";
```
lukesneeringer: Why this change? We do not have a port in the annotations, so this is needed, yes?

andreamlin:
Closing and opening https://github.com/googleapis/gapic-generator/pull/2350 to actually use the port specified in the service config.

PR #31, opened 2018-10-14
Refactor Wait to be a LRO and add Operations Service

landrito:
This PR has grown to be quite large 😕. LMK if you would like me to split this.


server/operations_service.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func (s *operationsServerImpl) GetOperation(ctx context.Context, in *lropb.GetOperationRequest) (*lropb.Operation, error) {
+	if strings.HasPrefix(in.Name, "operations/google.showcase.v1alpha2.Echo/Wait/") {
+		return s.handleWait(in)
+	}
+	return nil, status.Errorf(codes.NotFound, "Operation %q not found.", in.Name)
+}
+
+func (s *operationsServerImpl) handleWait(in *lropb.GetOperationRequest) (*lropb.Operation, error) {
```
pongad: I actually think it's more clear to inline this into `GetOperation`. Then you can do something like

```go
const prefix = "operations/.../Wait/"
if !strings.HasPrefix(in.Name, prefix) {error}
encodedBytes := in.Name[len(prefix):]
```

server/operations_service.go
```diff
@@ -0,0 +1,69 @@
...
+	waitReq := &pb.WaitRequest{}
+	encodedBytes := strings.TrimPrefix(in.Name, "operations/google.showcase.v1alpha2.Echo/Wait/")
+	waitReqBytes, _ := base64.StdEncoding.DecodeString(encodedBytes)
```
pongad: I think we need to check for this error.

server/waiter.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/golang/protobuf/ptypes"
+	pb "github.com/googleapis/gapic-showcase/server/genproto"
+	lropb "google.golang.org/genproto/googleapis/longrunning"
+)
+
+var waiterSingleton Waiter = &waiterImpl{
+	nowF: time.Now,
+}
+
+func GetWaiterSingleton() Waiter {
```
pongad: There's generally no need to do this. Just let others access the variable.

server/operations_service.go
landrito: Hm. I think I was putting the horse before the cart in this code decision. I was thinking this would be cleaner when we had multiple long running operations in showcase. I think going with what you have is cleaner until we get there and we can cross that design bridge then.

pongad: I support the decision of deferring this :D

server/echo_service_test.go
```diff
@@ -450,3 +373,12 @@ func TestPagedExpand(t *testing.T) {
 		}
 	}
 }
+
+func TestWait(t *testing.T) {
+	waiter := &mockWaiter{called: false}
+	server := &echoServerImpl{waiter: waiter}
+	server.Wait(context.Background(), nil)
+	if !waiter.called {
```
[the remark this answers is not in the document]
landrito: I'm not exactly following. Can you give an example?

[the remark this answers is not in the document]
pongad: Since `WaitRequest` has the time it should wait until, I thought it might be useful to record, something like

```go
type mockWaiter struct {
  req *pb.WaitRequest
}

func (w *mockWaiter) Wait(req *pb.WaitRequest) {w.req = req}
```

cmd/echo_service.go
```diff
@@ -198,6 +200,39 @@ func init() {
 		"The page size to send with this request")
 	commands = append(commands, pagedExpandCmd)
 
+	waitCmd := &cobra.Command{
+		Use:    "wait --seconds [amount] [content]",
```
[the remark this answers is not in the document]
landrito: It can! Done!

server/echo_service_test.go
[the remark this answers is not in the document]
landrito: Ah yeah. Done!

server/operations_service.go
landrito: Fixed!

server/operations_service.go
landrito: Done!

server/waiter.go
landrito: Fixed!

landrito:
Thanks for the review.

I think the coverage report is happening here right?

https://github.com/googleapis/gapic-showcase/blob/master/.circleci/config.yml#L53

pongad:
Oh I was confused by codecov saying there's no report. There's no report because build didn't succeed. My mistake!

PR #39, opened 2018-10-22
Introduce Messaging/Identity Service Protos

landrito:
## TODO
- [x] Add documentation


schema/chat_room.proto
```diff
@@ -0,0 +1,343 @@
...
+  rpc ListMessages(ListMessagesRequest) returns (ListMessagesResponse) {
+    option (google.api.http) = {
+      get: "/v1alpha3/{parent=rooms/*}/messages"
+      additional_bindings: {
+        get: "/v1alpha3/{parent=global}/messages"
+      }
+    };
+  }
+
+  rpc StreamMessages(stream CreateMessageRequest) returns (google.protobuf.Empty) {
```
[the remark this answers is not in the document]
landrito: These are covered elsewhere in the service. I tried organizing the methods by which resource they are tied to. With the unary stream being tied to a room and the stream stream covering both (so I just put it at the bottom).

Unary-Stream: https://github.com/googleapis/gapic-showcase/blob/chat/schema/chat_room.proto#L81

Stream-Stream: https://github.com/googleapis/gapic-showcase/blob/chat/schema/chat_room.proto#L168

schema/chat_room.proto
```diff
@@ -0,0 +1,343 @@
...
+      fields: ["query"]
+    };
+  }
+
+  rpc CreateMessage(CreateMessageRequest) returns (Message) {
+    option (google.api.http) = {
+      post: "/v1alpha3/{parent=rooms/*}/messages"
+      body: "*"
+      additional_bindings: {
+        post: "/v1alpha3/{parent=global}/messages",
```
noahdietz: Curious, in what other situation would a message be created where it's not in a room? I can think of DMs (between two people) or a group message thats not an explicitly named Room (like "API Client Tools Team"), but I'd say that such "containers" could be represented as a Room with the display_name being "Private Chat ndietz & landrito" or something.

schema/chat_room.proto
```diff
@@ -0,0 +1,343 @@
...
+  rpc Chat(stream CreateMessageRequest) returns (stream Message);
+}
+
+message Room {
+  string name = 1 [
+    (google.api.resource_set) = {
+      base_name: "Room"
+      resources: [
+        {base_name: "UserRoom"    path: "rooms/*"},
+        {base_name: "GlobalRoom"  path: "global"}
```
noahdietz: I'm kind of confused of the difference between a UserRoom and a GlobalRoom. Could you elaborate? Thank you :)

landrito: UserRoom is a user created room using the CreateRoom method. Global room is simply a default, explicitly named room singleton for the sole purpose of exercising resource sets. Admittedly there is probably a better way we could get the resource set coverage.

schema/chat_room.proto
landrito: I think this is a good idea but outside of scope of gapic-showcase. Bear in mind we are trying to minimize complexity whilst still having full test spec coverage.

schema/chat_room.proto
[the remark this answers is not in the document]
noahdietz: oh sorry I didn't even see the lines below, i was skimming to quickly. Apologies!!

I have no preference on grouping of streaming operations, I defer to what y'all feel is best

schema/chat_room.proto
noahdietz: ack. thanks!

schema/chat_room.proto
noahdietz: ack. agreed on simplicity, thanks for the reply.

schema/chat_room.proto
```diff
@@ -0,0 +1,343 @@
...
+    };
+  }
+
+  rpc ListRooms(ListRoomsRequest) returns (ListRoomsResponse) {
+    option (google.api.http) = {
+      get: "/v1alpha3/rooms"
+    };
+  }
+
+  rpc StreamRoom(StreamRoomRequest) returns (stream Message) {
```
vchudnov-g: This returns in a streaming fashion the list of rooms, right? I would suggest tweaking the name. When I first read this I thought it was going to stream the messages in a particular room.

schema/chat_room.proto
vchudnov-g: To clarify the semantics: What does `StreamMessages` stream to the server?  A series of messages that were queued up because the user was offline? (Trying to see the real-world use case...)

How about streaming an image file being uploaded by the client to the server?

schema/chat_room.proto
```diff
@@ -0,0 +1,343 @@
...
+      additional_bindings: {
+        post: "/v1alpha3/{parent=global}:streamMessages",
+        body: "*"
+      }
+    };
+  }
+
+  rpc Chat(stream CreateMessageRequest) returns (stream Message);
+}
+
```
vchudnov-g: To clarify the semantics, does `Chat` stream previous messages or only future messages (from either side, obviously)?

schema/chat_room.proto
```diff
@@ -0,0 +1,343 @@
...
+option go_package = "github.com/googleapis/gapic-showcase/server/genproto";
+option java_package = "com.google.showcase.v1alpha3";
+option java_multiple_files = true;
+
+option (google.api.metadata) = {
+  product_uri: "https://github.com/googleapis/gapic-showcase"
+};
+
+service ChatRoom {
+  rpc CreateRoom(CreateRoomRequest) returns (Room) {
```
alexander-fenster: It will much more resemble real APIs (such as Pub/Sub, Spanner) if the room creation returned `google.longrunning.Operation` and was asyncronous. After all, it's logical since room creation might call some long running operation under the hood.

alexander-fenster:
I feel the semantics of a chat resembles semantics of Pub/Sub so I'd vote for the RPCs to be similar to Pub/Sub in terms of their return types.

schema/chat_room.proto
alexander-fenster: (Luke just pointed out that Pub/Sub RPCs don't return LRO, but still, for me, room creation can be LRO)

schema/chat_room.proto
landrito: >  I thought it was going to stream the messages in a particular room.

This is the intended behavior. Hm. Maybe `StreamRoomMessages` is a better alternatives.

schema/chat_room.proto
landrito: I think for the purposes of gapic-showcase streaming only future messages would be sufficient.

vchudnov-g: Sounds reasonable. Let's comment that.

schema/chat_room.proto
vchudnov-g: Oh, I see (I overlooked the return value here). Yeah, I like `StreamRoomMessages` (and clarify they're past or future messages--not sure which you intend; past would make sense to complement the functionality of `Chat` below).

schema/chat_room.proto
landrito: Will do! Thanks for looking!

landrito:
@alexander-fenster I agree that the semantics resembles Pub/Sub. I think if this API were outside the scope of gapic-showcase, I'd agree to make it more consistent with pub/sub. I think having the resources follow this [section](https://cloud.google.com/apis/design/standard_methods) of the api design guide makes more sense in the overall picture of showcase since this API is trying to be a more standard representation of APIs built around resources.

Furthermore Pubsub is trying to solve a more general problem than what this chat service is trying express. That generality adds some cognitive overhead in understanding the API.

schema/chat_room.proto
landrito: I think this conversation boils down to how https://github.com/googleapis/gapic-showcase/pull/39#issuecomment-432024845 will be resolved.

schema/chat_room.proto
landrito: Hm. I think streaming an Image file could be a good alternative, let me try and improve upon what I have, before switching to a method with a different behavior.

schema/chat_room.proto
lukesneeringer: I would say ~90%-95% of our create methods return the resource (and that is [our expected starting point](https://cloud.google.com/apis/design/standard_methods#create)). We do sometimes have creates that return LROs, so I am happy to have one if it is appropriate.

Given that "create room" is likely to be the first thing some people do, I expect this is the wrong place to introduce an LRO.

schema/chat_room.proto
```diff
@@ -0,0 +1,343 @@
...
+      fields: ["name"]
+    };
+  }
+
+  rpc UpdateRoom(UpdateRoomRequest) returns (Room) {
+    option (google.api.http) = {
+      patch: "/v1alpha3/{name=rooms/*}"
+      body: "*"
+    };
+    option (google.api.method_signature) = {
```
lukesneeringer: Note: googleapis/api-common-protos#20 changee `google.api.method_signature` to `repeated` (the _status quo ante_ has been a consistent point of confusion).

schema/chat_room.proto
lukesneeringer: `StreamRoomMessages`, please -- the noun is `messages`.

schema/chat_room.proto
```diff
@@ -0,0 +1,343 @@
...
+
+  rpc ListRooms(ListRoomsRequest) returns (ListRoomsResponse) {
+    option (google.api.http) = {
+      get: "/v1alpha3/rooms"
+    };
+  }
+
+  rpc StreamRoom(StreamRoomRequest) returns (stream Message) {
+    option (google.api.http) = {
+      post: "/v1alpha3/{name=rooms/*}:streamRoom"
```
lukesneeringer: Similarly here, this should be:

```proto
option (google.api.http) = {
  post: "/v1alpha3/{name=rooms/*}/messages:stream"
};
```

The noun here is _messages_, not room.

schema/chat_room.proto
```diff
@@ -0,0 +1,343 @@
...
+      body: "*"
+      additional_bindings: {
+        post: "/v1alpha3/{name=global}:streamRoom"
```
lukesneeringer: This is wrong. It should be `/v1alpha3/rooms/-/messages:stream`, which will be covered by the original URI.

Context: https://cloud.google.com/apis/design/design_patterns#list_sub-collections

schema/chat_room.proto
```diff
@@ -0,0 +1,343 @@
...
+        body: "*"
+      }
+    };
+  }
+
+  rpc SearchAll(SearchAllRequest) returns (google.longrunning.Operation) {
```
lukesneeringer: Search all...what? Needs a noun. (It probably also does not need "all".)

Assuming this does what I think you intend it to do, it should be `rpc SearchMessages`.

schema/chat_room.proto
```diff
@@ -0,0 +1,343 @@
...
+    option (google.api.http) = {
+      post: "/v1alpha3:searchAll"
```
lukesneeringer: Most likely you want `/v1alpha3/rooms/-/messages:search` or perhaps `/v1alpha3/messages:search`.

schema/chat_room.proto
lukesneeringer: Still use `rooms/-` for this case (the creation equivalent of [this guidance](https://cloud.google.com/apis/design/design_patterns#get_unique_resource_from_sub-collection)).

(This applies throughout the PR; not noting each one.)

schema/chat_room.proto
```diff
@@ -0,0 +1,343 @@
...
+      }
+    };
+    option (google.api.method_signature) = {
+      fields: ["parent", "message"]
+    };
+  }
+
+  rpc GetMessage(GetMessageRequest) returns (Message) {
+    option (google.api.http) = {
+      get: "/v1alpha3/{parent=rooms/*}/{name=messages/*}"
```
lukesneeringer:
```proto
option (google.api.http) = {
  get: "/v1alpha3/{name=rooms/*/messages/*}"
};
```

And the `Message` message (har har) should not have a `parent` field at all.

(This applies throughout the standard methods for `Message`; not noting each one.)

schema/chat_room.proto
lukesneeringer: `StreamMessages` makes me expect to stream messages _down_. I might suggest something like "send messages"

I agree with @vchudnov-g that this probably would not be a real RPC if we actually wanted to run this service for more than a testbed, but I do not think it matters provided the RPC has an intuitive name. What matters is whether it will cause cognitive overhead for a reader.

If you do want to tweak it, perhaps streaming _audio_ would be a more sensible example? (e.g.  you say something into the API and it transcribes it) Images are too small to need streaming the bulk of the time.

lukesneeringer: On a separate note, I do not think `google.protobuf.Empty` is an appropriate result message here.

schema/chat_room.proto
lukesneeringer: Thoughts on `Connect` instead of `Chat`?

schema/chat_room.proto
lukesneeringer: I am fine with having a global room, but I very much do not want to have the `global` string like this. _Even if it happens to work_, this is not a supported use case of resource paths.

schema/chat_room.proto
```diff
@@ -0,0 +1,343 @@
...
+      }
+    };
+  }
+
+  rpc Chat(stream CreateMessageRequest) returns (stream Message);
+}
+
+message Room {
+  string name = 1 [
+    (google.api.resource_set) = {
```
lukesneeringer: See googleapis/api-common-protos#20 for some last minute changes to this annotation (sorry!).

schema/chat_room.proto
```diff
@@ -0,0 +1,343 @@
...
+}
+
+message CreateRoomRequest {
+  Room room = 1;
+}
+
+message GetRoomRequest {
+  string name = 1 [
+    (google.api.resource_type) = "google.showcase.v1alpha3.Room",
+    (google.api.required) = true
```
lukesneeringer: Here also, see googleapis/api-common-protos#20 for a change (sorry!).

schema/chat_room.proto
```diff
@@ -0,0 +1,343 @@
...
+  ];
+}
+
+message UpdateRoomRequest {
```
lukesneeringer: This does not follow [the semantics for update](https://cloud.google.com/apis/design/standard_methods#update).

It should be:

```proto
message UpdateRoomRequest {
  Room room = 1;
  google.protobuf.FieldMask update_mask = 2;
}
```

schema/chat_room.proto
```diff
@@ -0,0 +1,343 @@
...
+      get: "/v1alpha3/{parent=rooms/*}/{name=messages/*}"
+      additional_bindings: {
+        get: "/v1alpha3/{parent=global}/name={messages/*}"
+      }
+    };
+  }
+
+  rpc UpdateMessage(UpdateMessageRequest) returns (Message) {
+    option (google.api.http) = {
+      patch: "/v1alpha3/{parent=rooms/*}/{name=messages/*}"
```
lukesneeringer: This one has a slightly different correct answer:

```proto
option (google.api.http) = {
  patch: "/v1alpha3/{room.name=rooms/*/messages/*}"
};
```

Note the use of `room.name` instead of just `name`, which corresponds to `UpdateRoomRequest`'s only having `Room room`, and `name` is a sub-field of `room`.

schema/chat_room.proto
```diff
@@ -0,0 +1,343 @@
...
+  string parent = 1 [
+    (google.api.resource_type) = "google.showcase.v1alpha3.Room"];
+
+  string name = 2 [(google.api.resource).path = "messages/*"];
+
+  // This will eventually become a User message that will cover add coverage for
+  // shared resources.
+  string author = 3;
+
+  string content = 4;
```
lukesneeringer: Perhaps wrap `content` and `image` in a oneof? I have not seen any oneofs yet.

schema/chat_room.proto
```diff
@@ -0,0 +1,343 @@
...
+message SearchAllMetadata {
+  google.rpc.RetryInfo retry_info = 1;
+}
+
+message SearchAllResponse {
+  repeated Message messages = 1;
+}
+
+message Message {
+  string parent = 1 [
```
lukesneeringer: This field should not exist (see the RPCs for the explanation).

schema/chat_room.proto
```diff
@@ -0,0 +1,343 @@
...
+    (google.api.resource_type) = "google.showcase.v1alpha3.Room"];
+
+  string name = 2 [(google.api.resource).path = "messages/*"];
+
+  // This will eventually become a User message that will cover add coverage for
+  // shared resources.
+  string author = 3;
```
lukesneeringer: This is required, yeah?

schema/chat_room.proto
```diff
@@ -0,0 +1,343 @@
...
+  ];
+
+  string name = 2 [
+    (google.api.resource_type) = "google.showcase.v1alpha3.Message",
+    (google.api.required) = true
+  ];
+}
+
+message UpdateMessageRequest {
+  string parent = 1 [
```
lukesneeringer: Omit.

schema/chat_room.proto
```diff
@@ -0,0 +1,343 @@
...
+    (google.api.resource_type) = "google.showcase.v1alpha3.Room",
+    (google.api.required) = true
+  ];
+
+  Message message = 2;
+}
+
+message GetMessageRequest {
+  string parent = 1 [
```
lukesneeringer: Omit.

schema/chat_room.proto
```diff
@@ -0,0 +1,343 @@
...
+  string name = 2 [
+    (google.api.resource_type) = "google.showcase.v1alpha3.Message",
+    (google.api.required) = true
+  ];
+
+  Message message = 3;
+}
+
+message DeleteMessageRequest {
+  string parent = 1 [
```
lukesneeringer: Omit.

schema/chat_room.proto
```diff
@@ -0,0 +1,343 @@
...
+  ];
+}
+
+message ListMessagesRequest {
+  string parent = 1 [
+    (google.api.resource_type) = "google.showcase.v1alpha3.Room",
+    (google.api.required) = true
+  ];
+
+  string name = 2 [
```
lukesneeringer: Omit (note -- it is `name` that should go away this time, not `parent`).

schema/chat_room.proto
```diff
@@ -0,0 +1,343 @@
...
+    (google.api.resource_type) = "google.showcase.v1alpha3.Message",
+    (google.api.required) = true
+  ];
+
+  int32 page_size = 3;
+
+  string page_token = 4;
+}
```
lukesneeringer: Perhaps add `string filter` in a future version?

schema/chat_room.proto
```diff
@@ -0,0 +1,343 @@
...
+message ListMessagesResponse {
+  repeated Message messages = 1;
+
+  string next_page_token = 2;
+}
+
+message ChatroomBehaviorHeader {
+  enum ServerBehavior {
+    BEHAVIOR_UNSPECIFIED = 0;
+    DEFAULT_BEHAVIOR = 1;
```
lukesneeringer: Just `DEFAULT` and `STATELESS` is sufficient.

schema/chat_room.proto
```diff
@@ -0,0 +1,343 @@
...
+  string page_token = 4;
+}
+
+message ListMessagesResponse {
+  repeated Message messages = 1;
+
+  string next_page_token = 2;
+}
+
+message ChatroomBehaviorHeader {
```
lukesneeringer: What is this thing?

schema/chat_room.proto
landrito: fixed!

schema/chat_room.proto
landrito: It's better than Chat. We might be able to get something better than that though.

schema/chat_room.proto
landrito: I went with `StreamMessages`.

schema/chat_room.proto
landrito: I changed it to `SendMessages`

schema/chat_room.proto
landrito: Changed to be `StreamMessages`.

schema/chat_room.proto
landrito: I removed `global` from the proto altogether.

schema/chat_room.proto
landrito: Changed to SearchMessages

schema/chat_room.proto
landrito: Changed to /v1alpha3/rooms/-/messages:search

schema/chat_room.proto
landrito: Done. I was trying to figure out how to have the resource_set use the /global as a separate path. But that made everything terrible!

schema/chat_room.proto
landrito: Fixed to be `SendMessagesResponse`

schema/chat_room.proto
landrito: Done!

schema/chat_room.proto
landrito: Done!

schema/chat_room.proto
landrito: Done!

schema/chat_room.proto
landrito: Makes sense. Done!

schema/chat_room.proto
landrito: Done!

schema/chat_room.proto
landrito: Done!

schema/chat_room.proto
landrito: Done!

schema/chat_room.proto
landrito: Done!

schema/chat_room.proto
landrito: Done!

schema/chat_room.proto
landrito: Done!

schema/chat_room.proto
landrito: Done!

schema/chat_room.proto
landrito: Sure. I'll leave this out in this iteration.

schema/chat_room.proto
landrito: Done.

schema/chat_room.proto
landrito: This is an idea that @pongad tossed around in one of our testing meetings. This message will be serialized and passed in as metadata alongside any of the requests. If `DEFAULT` or `BEHAVIOR_UNSPECIFIED` or not passed, the showcase server will work as expected. The server will keep track of all messages, rooms and messages created.

If `STATELESS` is passed, the server will have specific expected request values and responses. RPCs will work with correct values but will not have the side-effect of changing state. ie The Server will always have the same rooms and messages. This will make it such that a test that uses a specific method, will not need any setup code. and the tests written on top of it will not have to worry about parallel tests causing an invalid state in the test environment. 

After writing this paragraph, I am on the 80% sure that this is unnecessary to be honest. I am not really thinking that there is gain to having a stateless solution.

lukesneeringer: Yeah, this sounds like YAGNI to me (and confusing YAGNI at that).
I _think_ what you are trying to get here is the ability to not have to do "build up" in order to run specific tests? I assert you could get 90% of that effect by just having a documented shortcut, e.g. "If you start Showcase with the `--create-scaffolding` switch, it automatically creates three rooms for you."

(And you might not even care about _that._)

schema/chat_room.proto
```diff
@@ -0,0 +1,288 @@
...
+      fields: ["name"]
+    };
+  }
+
+  rpc UpdateRoom(UpdateRoomRequest) returns (Room) {
+    option (google.api.http) = {
+      patch: "/v1alpha3/{room.name=rooms/*}"
+      body: "*"
+    };
+    option (google.api.method_signature) = {
```
lukesneeringer: Hmm, how will these method signatures interact with field masks? It seems like if I call `client.update_room(name='foobar')`, the actual behavior is going to be to no-op, because I did not set a field mask, and we have no automated-field-mask-setting functionality here.

(Which is frustrating, because I get that your goal here is to use nested fields, which you effectively can not do.)

schema/chat_room.proto
```diff
@@ -0,0 +1,288 @@
...
+      fields: ["name"]
+    };
+  }
+
+  rpc ListRooms(ListRoomsRequest) returns (ListRoomsResponse) {
+    option (google.api.http) = {
+      get: "/v1alpha3/rooms"
+    };
```
lukesneeringer: This is going to end up being a bit of a weird case, because the common need here is to send a request with literally nothing in it, but I expect most generators still make you instantiate something (or send empty dict or whatnot).

The change to `google.api.method_signature` also effectively precludes our ability to detect an _empty_ signature (because both segments of the repeated->repeated line here lack presence testing). We could potentially augment this in the future by having a `bool empty`.

Bummer. :-/

schema/chat_room.proto
```diff
@@ -0,0 +1,288 @@
...
+    option (google.api.method_signature) = {
+      fields: ["parent", "message"]
+    };
+  }
+
+  rpc GetMessage(GetMessageRequest) returns (Message) {
+    option (google.api.http) = {
+      get: "/v1alpha3/{name=rooms/*/messages/*}"
+    };
```
lukesneeringer: No `google.api.method_signature` here?

```suggestion
    };
    option (google.api.method_signature) = {
      fields: ["name"]
    };
```

schema/chat_room.proto
lukesneeringer: Thinking out loud: Maybe not. Many services consider a lack of a field mask to be the equivalent of `"*"`, and this one could too.

schema/chat_room.proto
```diff
@@ -0,0 +1,288 @@
...
+    option (google.api.http) = {
+      patch: "/v1alpha3/{message.name=rooms/*/messages/*}"
+      body: "*"
+    };
+  }
+
+  rpc DeleteMessage(DeleteMessageRequest) returns (google.protobuf.Empty) {
+    option (google.api.http) = {
+      delete: "/v1alpha3/{name=rooms/*/messages/*}"
+    };
```
lukesneeringer:
```suggestion
    };
    option (google.api.method_signature) = {
      fields: ["name"]
    };
```

schema/chat_room.proto
```diff
@@ -0,0 +1,288 @@
...
+  }
+
+  rpc ListMessages(ListMessagesRequest) returns (ListMessagesResponse) {
+    option (google.api.http) = {
+      get: "/v1alpha3/{parent=rooms/*}/messages"
+    };
```
lukesneeringer:
```suggestion
    };
    option (google.api.method_signature) = {
      fields: ["parent"]
    };
```

schema/chat_room.proto
```diff
@@ -0,0 +1,288 @@
...
+  }
+
+  rpc SearchMessages(SearchMessagesRequest) returns (google.longrunning.Operation) {
+    option (google.api.http) = {
+      post: "/v1alpha3/rooms/-/messages:search"
```
lukesneeringer: Would it make sense to _allow_ searching within a particular room? (e.g. `"/v1alpha3/{parent=rooms/*}/messages:search"`)

lukesneeringer:
On resource sets: The clearest thing I can think of to add to make resource sets work would be direct messages, which then means that messages can be owned by either a room or by a direct message collection resource.

schema/chat_room.proto
landrito: I'd consider lack of field mask to be equivalent to "*".

schema/chat_room.proto
landrito: That definitely makes sense, but I think that making the search space smaller makes it less convincing that this is a LRO.

schema/chat_room.proto
lukesneeringer: Cool, great. Sorry for the fire drill.

schema/chat_room.proto
lukesneeringer: But it is an LRO if _any_ request that comes from it would take time.
(Also, have you ever searched messages in Slack when there is a multi-year history. Fear not, it is totally reasonable for this to be an LRO ;-))

schema/chat_room.proto
landrito: Ah that's a good point. I assume we are too late to get empty into https://github.com/googleapis/api-common-protos/pull/20

landrito:
This is ready for another look!

.circleci/config.yml
```diff
@@ -80,7 +80,7 @@ jobs:
       - run:
           name: Stage showcase in proto dependencies
           command: |
-              git clone -b input-contract https://github.com/googleapis/api-common-protos.git
+              git clone -b input-contract-edits https://github.com/googleapis/api-common-protos.git
```
lukesneeringer: * TODO(me): Try to merge googleapis/api-common-protos#20 before this goes in.
  * TODO(landrito): Revert this ones it that is done.

```suggestion
              git clone -b input-contract https://github.com/googleapis/api-common-protos.git
```

README.md
```diff
@@ -189,8 +189,8 @@ $ gapic-showcase echo hello world
 ### Staged Protocol Buffer Files
 The [proto files](schema/) found in the gapic-showcase repository are not compilable in
 isolation. This is to avoid duplication of the protofiles that showcase depends
-on, namely the API and API client configurations found in the `input-contract` branch of
-[api-common-protos](https://github.com/googleapis/api-common-protos/tree/input-contract).
+on, namely the API and API client configurations found in the `input-contract-edits` branch of
```
lukesneeringer: _Ibid._

schema/messaging.proto
```diff
@@ -0,0 +1,508 @@
...
+option java_multiple_files = true;
+
+option (google.api.metadata) = {
+  product_uri: "https://github.com/googleapis/gapic-showcase"
+};
+
+// This resource is defined in google/showcase/v1alpha3/user.proto. Since
+// google/showcase/v1alpha3/user.proto is not imported from this file, the
+// resource needs to be defined here to be used in this API.
+option (google.api.resource_definition) = {
```
lukesneeringer: Oh, nice. This is a surprise benefit of @chrisdunelm's work that I had not considered -- stubbed resources (which are totally a thing).

schema/messaging.proto
```diff
@@ -0,0 +1,508 @@
...
+  path: "users/{user_id}"
+};
+
+// A resource set defining the possible parent paths of a
+// message.
+option (google.api.resource_set_definition) = {
+  name: "MessageParent"
+  resources: [{
+    name: "Profile",
+    path: "users/{user_id}/profile"
```
lukesneeringer: @chrisdunelm @andreamlin Need to make sure your implementations support the singleton pattern. (This does show up in real APIs.) I expect they do, but worth explicitly checking.

schema/messaging.proto
```diff
@@ -0,0 +1,508 @@
...
+    "google.showcase.v1alpha3.Room"
+  ]
+};
+
+// A simple messaging service that implements chat rooms and profile posts.
+//
+// This messaging service showcases the features that API clients
+// generated by gapic-generators implement.
+service Messaging {
+  // Creates a room. The fields "room.name", "room.create_time", and
```
lukesneeringer: Just `Creates a room.` here; we document the rest on the fields themselves. (I will note it there.)

schema/messaging.proto
```diff
@@ -0,0 +1,508 @@
...
+
+    // This option signifies that the resulting generated client could include
+    // a method overload which accepts the following listed fields as
+    // parameters.
+    option (google.api.method_signature) = {
+      fields: ["room.display_name", "room.description"]
+    };
+  }
+
+  // Retrieves the Room with the given uri.
```
lukesneeringer: Nitpick:

```suggestion
  // Retrieves the Room with the given URI.
```

schema/messaging.proto
```diff
@@ -0,0 +1,508 @@
...
+
+  // Updates a room.
+  rpc UpdateRoom(UpdateRoomRequest) returns (Room) {
+    option (google.api.http) = {
+      patch: "/v1alpha3/{room.name=rooms/*}"
+      body: "*"
+    };
+  }
+
+  // Deletes a room and all of it's messages.
```
lukesneeringer: Nit: 

```suggestion
  // Deletes a room and all of its messages.
```

schema/messaging.proto
```diff
@@ -0,0 +1,508 @@
...
+service Messaging {
+  // Creates a room. The fields "room.name", "room.create_time", and
+  // "room.update_time" will be ignored if passed to this method.
+  rpc CreateRoom(CreateRoomRequest) returns (Room) {
+    option (google.api.http) = {
+      post: "/v1alpha3/rooms"
+      body: "*"
+    };
+
+    // This option signifies that the resulting generated client could include
```
lukesneeringer: Thinking out loud; should we use the `(--` syntax here to distinguish from what you would actually want to end up in user documentation?

(No strong opinion.)

schema/messaging.proto
```diff
@@ -0,0 +1,508 @@
...
+  rpc ListRooms(ListRoomsRequest) returns (ListRoomsResponse) {
+    option (google.api.http) = {
+      get: "/v1alpha3/rooms"
+    };
+  }
+
+  // Creates a message. If the parent field corresponds to a RoomMessage
+  // resource, the message is understood to be a message in a chatroom. If the
+  // parent field is a ProfileMessage resource name, the message is understood
+  // to be a post on a user profile. The fields "message.name",
```
lukesneeringer: Omit this last sentence (as discussed above).

schema/messaging.proto
```diff
@@ -0,0 +1,508 @@
...
+    option (google.api.http) = {
+      post: "/v1alpha3/{parent=rooms/*}/messages"
+      body: "*"
+      additional_bindings: {
+        post: "v1alpha3/{parent=users/*/profile}/messages"
+        body: "*"
+      }
+    };
+
+    // This option signifies that the resulting generated client could include
```
lukesneeringer: I liked this comment the first time I saw it, but it is distracting at this point and actually makes the proto _harder_ to read. Perhaps that is because I read protos all day?

Maybe show this comment once? Or have a large, explanatory comment at the top? (I think that might work the best, but unsure.)

.circleci/config.yml
landrito: Will Do!

schema/messaging.proto
landrito: Agreed!

schema/messaging.proto
landrito: Done!

schema/messaging.proto
landrito: Done everywhere.

schema/messaging.proto
landrito: Done everywhere.

schema/messaging.proto
landrito: Removed per comment below.

schema/messaging.proto
landrito: Done.

schema/messaging.proto
landrito: Removed. I think the documentation of these features will be better elsewhere in showcase. It will be explicitly pointed out in the showcase spec and will link to the gapic-generator standard once thats open sourced.

schema/messaging.proto
```diff
@@ -358,7 +345,16 @@ message CreateMessageRequest {
   // The URI of the chat room or user profile that this message will be tied
   // to.
   string parent = 1 [
-    (google.api.resource_reference) = "google.showcase.v1alpha3.MessageParent",
+    (google.api.resource_set) = {
```
landrito: @lukesneeringer, @chrisdunelm Which of these do you think makes more sense stylistically?

schema/messaging.proto
lukesneeringer: I buy that.

schema/messaging.proto
```diff
@@ -0,0 +1,463 @@
...
+    (google.api.field_behavior) = OUTPUT_ONLY
+  ];
+
+  // The latest timestamp at which the room was updated.
+  google.protobuf.Timestamp update_time = 5 [
+    (google.api.field_behavior) = OUTPUT_ONLY
+  ];
+}
+
+// The request message for the google.showcase.v1alpha3.Messaging\CreateRoom
```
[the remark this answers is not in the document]
landrito: This is the syntax of the method name that is passed to the interceptors in grpc-go, I'd think that it is established, but I'm unsure.

schema/messaging.proto
```diff
@@ -0,0 +1,463 @@
...
+  // A token to retrieve next page of results.
+  // Pass this value in ListRoomsRequest.page_token field in the subsequent
+  // call to `google.showcase.v1alpha3.Message\ListRooms` method to retrieve the
+  // next page of results.
+  string next_page_token = 2;
+}
+
+// This protocol buffer message represents a message sent to a chat room or
+// posted on a user profile.
+message Message {
```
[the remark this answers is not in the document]
landrito: I like `Blurb` especially since there's little naming overloads. Another possibility is `Statement` but that's also overloaded (albeit less overloaded than message). I'll try `Blurb`.

schema/user.proto
```diff
@@ -0,0 +1,167 @@
...
+  }
+}
+
+// A user.
+message User {
+  // The uri of the user.
+  string name = 1 [(google.api.resource).path = "users/{user_id}"];
+
+  // The username of the user.
+  string username = 2 [(google.api.field_behavior) = REQUIRED];
```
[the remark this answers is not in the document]
landrito: Done!

schema/user.proto
```diff
@@ -0,0 +1,167 @@
...
+  // The latest timestamp at which the user was updated.
+  google.protobuf.Timestamp update_time = 5 [
+    (google.api.field_behavior) = OUTPUT_ONLY
+  ];
+}
+
+// The request message for the google.showcase.v1alpha3.Identity\CreateUser
+// method.
+message CreateUserRequest {
+  // The user to be created. The "user.name", "user.create_time", and
```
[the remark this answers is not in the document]
landrito: Done!

schema/user.proto
```diff
@@ -0,0 +1,167 @@
...
+  // "user.update_time" fields will be ignored by the server in this request.
+  User user = 1;
+}
+
+// The request message for the google.showcase.v1alpha3.Identity\GetUser
+// method.
+message GetUserRequest {
+  // The uri of the requested user.
```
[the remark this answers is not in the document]
landrito: Done everywhere!

schema/user.proto
```diff
@@ -0,0 +1,167 @@
...
+
+  // The field mask to determine wich fields are to be updated. If empty, the
+  // server will assume all fields are to be updated.
+  google.protobuf.FieldMask update_mask = 2;
+}
+
+// The request message for the google.showcase.v1alpha3.Identity\DeleteUser
+// method.
+message DeleteUserRequest {
+  // The uri of the user to be deleted.
```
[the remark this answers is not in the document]
landrito: Done everywhere.

schema/user.proto
```diff
@@ -0,0 +1,167 @@
...
+}
+
+// The request message for the google.showcase.v1alpha3.Identity\ListUsers
+// method.
+message ListUsersRequest {
+  // The maximum number of users to be returned. Server may return fewer users
+  // than requested. If unspecified, server will pick an appropriate default.
+  int32 page_size = 1;
+
+  // Typically, this is the value of
```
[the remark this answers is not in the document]
landrito: Done everywhere.

schema/messaging.proto
```diff
@@ -0,0 +1,463 @@
...
+  // The latest timestamp at which the room was updated.
+  google.protobuf.Timestamp update_time = 5 [
+    (google.api.field_behavior) = OUTPUT_ONLY
+  ];
+}
+
+// The request message for the google.showcase.v1alpha3.Messaging\CreateRoom
+// method.
+message CreateRoomRequest {
+  // The room to be created. The "room.name", "room.create_time", and
```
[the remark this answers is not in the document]
landrito: Woops missed acouple. Done!

schema/messaging.proto
landrito: Nope. It is needed because `user.proto` is not imported from `messaging.proto` thus the resource definition is not guaranteed to be loaded from user.proto.

schema/messaging.proto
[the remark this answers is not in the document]
landrito: Done!

schema/messaging.proto
```diff
@@ -0,0 +1,463 @@
...
+  string name = 1 [
+    (google.api.resource_set) = {
+      resources: [
+        {name: "RoomMessage", path: "rooms/{room_id}/messages/{message_id}"},
+        {name: "ProfileMessage", path: "user/{user_id}/profile/messages/{message_id}"}
+      ]
+    }];
+
+  // The URI of the author of this method.
+  string author = 2 [
```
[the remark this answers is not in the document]
landrito: Done!

schema/messaging.proto
```diff
@@ -0,0 +1,463 @@
...
+  // The URI of the chat room.
+  string name = 1 [
+    (google.api.resource_set) = {
+      resources: [
+        {name: "RoomMessage", path: "rooms/{room_id}/messages/{message_id}"},
+        {name: "ProfileMessage", path: "user/{user_id}/profile/messages/{message_id}"}
+      ]
+    }];
+
+  // The URI of the author of this method.
```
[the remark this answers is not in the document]
landrito: Done!

schema/messaging.proto
landrito: Agreed. Fixed.

.circleci/config.yml
landrito: Done.

README.md
landrito: Done.

schema/messaging.proto
```diff
@@ -0,0 +1,463 @@
...
+      resource_references: [
+        "google.showcase.v1alpha3.Room"
+      ]
+    },
+    (google.api.field_behavior) = REQUIRED
+  ];
+
+  // The message to be created. The "message.name", "message.create_time", and
+  // "message.update_time" fields will be ignored by the server in this request.
+  Message message = 2;
```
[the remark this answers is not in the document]
landrito: Fixed!

schema/messaging.proto
```diff
@@ -0,0 +1,463 @@
...
+    (google.api.resource_reference) = "google.showcase.v1alpha3.Message",
+    (google.api.field_behavior) = REQUIRED
+  ];
+}
+
+// The request message for the google.showcase.v1alpha3.Messaging\UpdateMessage
+// method.
+message UpdateMessageRequest {
+  // The message to be updated.
+  Message message = 1;
```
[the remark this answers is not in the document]
landrito: Fixed!

schema/messaging.proto
```diff
@@ -0,0 +1,463 @@
...
+  // the next page of results.
+  string next_page_token = 2;
+}
+
+// The request message for the google.showcase.v1alpha3.Messaging\SearchMessages
+// method.
+message SearchMessagesRequest {
+  // The query used to search for messages containing to words of this string.
+  // Only posts that contain an exact match of a queried word will be returned.
+  string query = 1 [(google.api.field_behavior) = REQUIRED];
```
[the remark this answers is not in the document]
landrito: Added.

schema/messaging.proto
```diff
@@ -0,0 +1,463 @@
...
+message SearchMessagesMetadata {
+  // This signals to the client to when to next poll for response.
+  google.rpc.RetryInfo retry_info = 1;
+}
+
+// The operation response message for the
+// google.showcase.v1alpha3.Messaging\SearchMessages method.
+message SearchMessagesResponse {
+  // Messages that matched the search query.
+  repeated Message messages = 1;
```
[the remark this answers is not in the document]
landrito: Added.

schema/messaging.proto
```diff
@@ -0,0 +1,463 @@
...
+}
+
+// The request message for the google.showcase.v1alpha3.Messaging\StreamMessages
+// method.
+message StreamMessagesRequest {
+  // The URI of a chat room or user profile whos massages are to be streamed.
```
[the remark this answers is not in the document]
landrito: Done.

schema/messaging.proto
```diff
@@ -0,0 +1,463 @@
...
+  string name = 1 [
+    (google.api.resource_reference) = "google.showcase.v1alpha3.Room",
+    (google.api.field_behavior) = REQUIRED
+  ];
+}
+
+// The response message for the google.showcase.v1alpha3.Messaging\SendMessages
+// method.
+message SendMessagesResponse {}
```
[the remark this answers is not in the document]
landrito: Added.

schema/messaging.proto
lukesneeringer: Oh. Hmm. Interesting.

Why is user.proto not imported from messaging.proto? Is `Identity` not needed there?

schema/messaging.proto
```diff
@@ -0,0 +1,482 @@
...
+
+  // Lists all chat rooms.
+  rpc ListRooms(ListRoomsRequest) returns (ListRoomsResponse) {
+    option (google.api.http) = {
+      get: "/v1alpha3/rooms"
+    };
+  }
+
+  // Creates a blurb. If the parent field corresponds to a RoomBlurb
+  // resource, the blurb is understood to be a message in a chatroom. If the
```
lukesneeringer: Reword:

> If the parent is a room, the blurb is understood to be a message in that room.
> If the parent is a profile, the blurb is understood to be a post on the profile.

schema/messaging.proto
```diff
@@ -0,0 +1,482 @@
...
+  rpc CreateBlurb(CreateBlurbRequest) returns (Blurb) {
+    option (google.api.http) = {
+      post: "/v1alpha3/{parent=rooms/*}/blurbs"
+      body: "*"
+      additional_bindings: {
+        post: "v1alpha3/{parent=users/*/profile}/blurbs"
+        body: "*"
+      }
+    };
+    option (google.api.method_signature) = {
```
lukesneeringer: Optional: I think it would be preferable to put the things that do not change in front of the things that do (e.g. `blurb.user` before `blurb.text`/`blurb.image`).

schema/messaging.proto
```diff
@@ -0,0 +1,482 @@
...
+      fields: ["parent"]
+    };
+  }
+
+  // This method searches through all blurbs across all rooms and profiles
+  // for blurbs containing to words found in the query. Only posts that
+  // contain an exact match of a queried word will be returned.
+  rpc SearchBlurbs(SearchBlurbsRequest) returns (google.longrunning.Operation) {
+    option (google.api.http) = {
+      post: "/v1alpha3/rooms/-/blurbs:search"
```
lukesneeringer: Optional: I think this should be `{parent=rooms/-}`, to permit expansibility later.

schema/messaging.proto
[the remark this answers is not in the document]
lukesneeringer: I am too. Comments can always change later, so do not sweat it. :-)

schema/messaging.proto
```diff
@@ -0,0 +1,482 @@
...
+  ];
+}
+
+// The request message for the google.showcase.v1alpha3.Messaging\CreateBlurb
+// method.
+message CreateBlurbRequest {
+  // The resource name of the chat room or user profile that this blurb will
+  // be tied to.
+  string parent = 1 [
+    (google.api.resource_reference) = "BlurbParent"
```
lukesneeringer: I do not love this (spec-wise; you faithfully follow the spec here) but I have no better answer.

chrisdunelm: Luke: I'm probably missing context; what specifically do you dislike about this?

schema/messaging.proto
landrito: This is good feedback that shouldn't be optional but rather standardized (unless backwards incompatible of course).

schema/messaging.proto
landrito: Correct, Identity is not needed. But for actual usage of the api, the flow would be something like the following which is why I created the identity proto. 

```
user = Identity.CreateUser()
room = Messaging.CreateRoom()
messaging.CreateMessage(room, {user: user, text: text})

File: schema/messaging.proto (hunk 63 above)
Comment (landrito) [reply schema/messaging.proto]: I agree that I don't love this, but I don't hate it. It's definitely followable.

File: schema/messaging.proto (hunk 62 above)
Comment (landrito) [reply schema/messaging.proto]: Is it common add http params when they are not present in the request message?

File: schema/messaging.proto [hunk 64]
```diff
@@ -479,3 +479,24 @@ message SendBlurbsResponse {
...
 }
+
+// The request message for the google.showcase.v1alpha3.Messaging\Connect
+// method.
+message ConnectRequest {
+  message ConnectConfig {
+    // The room or profile to follow and create messages for.
+    string parent = 1 [
+      (google.api.resource_reference) = "BlurbParent",
+      (google.api.field_behavior) = REQUIRED
```
landrito: Does it make sense to put `REQUIRED` on a field that must be sent iff the parent message is used?

lukesneeringer: I do not understand the question. If parent can be empty, then it is not required. If parent must be non-empty, then it can be set as required.

schema/messaging.proto
lukesneeringer: We only use BlurbParent in one spot, ideally it should be declared nearer or inline. But that would either add _another_ permutation to the annotation (because this approach is still needed) or restrict functionality.

It is fine as-is, if not as ideal as I might hope.

schema/messaging.proto
lukesneeringer: Ah, I see.

Something to think about: I think it is reasonable to expect an entire package to be generated together (and for errors to happen if someone tries to generate only parts of a package). We should probably add a note about that in our documentation.

This is still fine as-is, however. _Stet._

schema/messaging.proto
lukesneeringer: But there are cases where it is backwards-incompatible, so sadly it can not be a hard-and-fast rule.

There are a couple of legitimate reasons, also -- pretend that one of the two second arguments was _always_ sent but often the third argument was omitted (even if it is common to the two signatures). Then the ordering you used would be appropriate.

TL;DR: The ordering was wrong here. It is not a _per sé_ error, however.

schema/messaging.proto
lukesneeringer: I am also suggesting that `string parent` should be added to the request message (if you do this).

schema/messaging.proto
```diff
@@ -0,0 +1,463 @@
...
+  // The query used to search for messages containing to words of this string.
+  // Only posts that contain an exact match of a queried word will be returned.
+  string query = 1 [(google.api.field_behavior) = REQUIRED];
+}
+
+// The operation metadata message for the
+// google.showcase.v1alpha3.Messaging\SearchMessages method.
+message SearchMessagesMetadata {
+  // This signals to the client to when to next poll for response.
+  google.rpc.RetryInfo retry_info = 1;
```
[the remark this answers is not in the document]
pongad: OT: I don't see this mentioned in API style guide. Should our generators make use of this?

schema/messaging.proto
landrito: Bah, `parent` is overloaded in this context. I mean to say: the parent _field_ must be set IFF ConnectConfig is used in the oneof below.  but if the ConnectConfig is not used then the parent _field_ is not required. 

In some contexts the parent field is required and in some contexts it is not. I'm guessing `(google.api.field_behavior) = REQUIRED` will be too strict fort his specific case though.

schema/messaging.proto
landrito: Ack.

schema/messaging.proto
landrito: Done.

schema/messaging.proto
landrito: Ah gotcha. SGTM.

schema/messaging.proto
lukesneeringer: Correct. You should not include the `field_behavior` annotation at all in that case.

schema/messaging.proto
landrito: > We only use BlurbParent in one spot. 

BlurbParent is used in two spots. IDK if that changes your feelings of this though.

schema/messaging.proto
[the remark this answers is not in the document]
landrito: +1. I don't think this message has wide adoption amongst google APIs but building a GAPIC feature on top of this would be great. 

The only API I can think of that is returning this message is Spanner by returning this as part of their response headers for certain methods. /shrug if it is documented well though.

PR #45, opened 2018-11-09
Identity Server Implemetation

landrito:


server/user_db.go
```diff
@@ -0,0 +1,238 @@
...
+			codes.InvalidArgument,
+			"The field `email` is required.")
+	}
+	return nil
+}
+
+func (db *userDb) anyUser(f func(*pb.User) bool) bool {
+	for _, entry := range db.users {
+		if entry.deleted {
+			continue
```
noahdietz: Considering there is no logic following the `if f()` check, might be cleaner to do 
```
if !entry.deleted && f(entry.user) {
  return true
}
```
Not blocking, just a suggestion!

server/identity_service.go
```diff
@@ -0,0 +1,55 @@
...
+func NewIdentityServer() pb.IdentityServer {
+	return &identityServerImpl{db: NewUserDb()}
+}
+
+type identityServerImpl struct {
+	db UserDb
+}
+
+// Creates a user.
+func (s *identityServerImpl) CreateUser(ctx context.Context, in *pb.CreateUserRequest) (*pb.User, error) {
```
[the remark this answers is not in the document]
landrito: The param is needed to fulfill the server interface, I changed the param name to `_`.

server/identity_service_test.go
```diff
@@ -0,0 +1,363 @@
...
+		})
+	if err != nil {
+		t.Errorf("Create: unexpected err %+v", err)
+	}
+
+	_, err = s.DeleteUser(
+		context.Background(),
+		&pb.DeleteUserRequest{Name: delete.Name})
+	if err != nil {
+		t.Errorf("Create: unexpected err %+v", err)
```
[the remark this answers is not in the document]
landrito: Done

server/identity_service_test.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+
+	got, err = s.GetUser(
+		context.Background(),
+		&pb.GetUserRequest{Name: updated.GetName()})
+	if err != nil {
+		t.Errorf("Get: unexpected err %+v", err)
+	}
+	if !proto.Equal(updated, got) {
+		t.Error("Expected to get created user.")
```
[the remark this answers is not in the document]
landrito: Done

server/identity_service_test.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+
+	r, err = s.ListUsers(
+		context.Background(),
+		&pb.ListUsersRequest{PageSize: 10, PageToken: r.GetNextPageToken()})
+	if len(r.GetUsers()) != 1 {
+		t.Errorf("List want: page size %d, got %d", 1, len(r.GetUsers()))
+	}
+	if !proto.Equal(got, r.GetUsers()[0]) {
+		t.Errorf("List want: first user %+v, got %+v", first, r.GetUsers()[0])
```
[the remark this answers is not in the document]
landrito: Done.

server/identity_service_test.go
```diff
@@ -0,0 +1,363 @@
...
+		t.Errorf("Delete: unexpected err %+v", err)
+	}
+
+	_, err = s.GetUser(
+		context.Background(),
+		&pb.GetUserRequest{Name: created.Name})
+	status, _ := status.FromError(err)
+	if status.Code() != codes.NotFound {
+		t.Errorf(
+			"Create: Want error code %d got %d",
```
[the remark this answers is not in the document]
landrito: Done

server/identity_service_test.go
```diff
@@ -0,0 +1,363 @@
...
+	_, err := s.UpdateUser(
+		context.Background(),
+		&pb.UpdateUserRequest{
+			User:       nil,
+			UpdateMask: &field_mask.FieldMask{Paths: []string{"email"}},
+		})
+	status, _ := status.FromError(err)
+	if status.Code() != codes.Unimplemented {
+		t.Errorf(
+			"Create: Want error code %d got %d",
```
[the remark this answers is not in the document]
landrito: Done!

server/user_db.go
landrito: lol woops! Fixed!

server/user_db.go
```diff
@@ -0,0 +1,235 @@
...
+
+	nextToken := ""
+	if start+offset < len(db.users) {
+		nextToken = db.token.ForIndex(start + offset)
+	}
+
+	return &pb.ListUsersResponse{Users: users, NextPageToken: nextToken}, nil
+}
+
+func (db *userDb) validate(u *pb.User) error {
```
pongad: I don't think this needs to be a member of userDb right?

server/unique_id.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"sync/atomic"
+)
+
+type uniqID struct {
+	i int64
+}
+
+func (u *uniqID) id() int64 {
```
pongad: The name `id` sounds like a property access, maybe `next`?

server/user_db.go
```diff
@@ -0,0 +1,235 @@
...
+			"A user with the email %s already exists.",
+			u.GetEmail())
+	}
+
+	// Assign info.
+	id := db.uid.id()
+	name := fmt.Sprintf("users/%d", id)
+	now := ptypes.TimestampNow()
+
+	u.Name = name
```
pongad: Is it an error if user provide name or time?

server/user_db.go
```diff
@@ -0,0 +1,235 @@
...
+	user    *pb.User
+	deleted bool
+}
+
+type dbIndex int
+
+type userDb struct {
+	uid   *uniqID
+	token TokenGenerator
+	mu    sync.Mutex
```
pongad: `mu` protects `keys` and `users` right? In that case, add new line before `mu`. Idiomatically mutex is the first member in the "group" it's protecting.

server/user_db.go
```diff
@@ -0,0 +1,235 @@
...
+	// Insert.
+	index := dbIndex(len(db.users))
+	db.users = append(db.users, userEntry{user: u, deleted: false})
+	db.keys[name] = index
+
+	return u, nil
+}
+
+func (db *userDb) Get(s string) (*pb.User, error) {
+	if v, ok := db.keys[s]; ok {
```
pongad: We need mutex lock here right?

server/user_db.go
```diff
@@ -0,0 +1,235 @@
...
+
+	db.mu.Lock()
+	defer db.mu.Unlock()
+
+	err := db.validate(u)
+	if err != nil {
+		return nil, err
+	}
+
+	if i, ok := db.keys[u.GetName()]; ok {
```
pongad: I'd encourage flipping the logic:
```
i, ok := db.keys[u.GetName()]
if !ok {
  return error
}
```

server/user_db.go
```diff
@@ -0,0 +1,235 @@
...
+		entry := db.users[int(i)]
+		if !entry.deleted {
+			// Validate Unique Fields.
+			hasSameName := func(x *pb.User) bool {
```
pongad: Maybe factor this into `uniqNameEmail`?

server/user_db.go
```diff
@@ -0,0 +1,235 @@
...
+	Create(u *pb.User) (*pb.User, error)
+	Get(name string) (*pb.User, error)
+	Update(u *pb.User, f *field_mask.FieldMask) (*pb.User, error)
+	Delete(name string) error
+	List(page_size int32, token string) (*pb.ListUsersResponse, error)
+}
+
+type userEntry struct {
+	user    *pb.User
+	deleted bool
```
pongad: If we keep adding and removing, we'll eventually run out of memory. Probably not a big consideration here though?

server/user_db.go
landrito: Correct. Fixed.

server/unique_id.go
landrito: Changed!

server/user_db.go
landrito: Nope they should just be ignored. I fixed the update implementation to follow that.

server/user_db.go
landrito: Thanks! Done!

server/user_db.go
landrito: Good Catch! I didn't need one before when I was using just a map, but when I switch to using a pseudo ordered map, I do. Fixed!

server/user_db.go
landrito: Done!

server/user_db.go
landrito: Done, but we do lose some granularity in the error message since we won't be able to specify which field had a collision.

server/user_db.go
landrito: Agreed. Not too big of a consideration. I made the server rely on some DB interface for this reason. If ever we need to scale, we can just write a userDB interface to an actual database. I foresee any tests ever getting near causing an OOM error if the server is only alive for the current testing session.

landrito: Furthermore, the OOM problem is inherent making an in-memory db. For this application the deleted bool is simply to provide a simple way of generating page tokens. Since nothing is ever removed from the list, we can rely on index for page tokens.

server/user_db.go
pongad: I don't have a strong opinion on this but something like this would also work

```
func (db *DB) VerifyNoDupe(name, email string) error {
  for _, u := range db.users {
    if u.Name == name {return errors.E("duplicate name")}
    if u.Email == email {return errors.E("duplicate email")}
  }
  return nil
}
```
then we don't need to deal with function pointers.

landrito: Yeah thats better! My javascript is showing 😁.

PR #46, opened 2018-11-20
Implement MessagingService Rooms

landrito:


server/messaging_service.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+type messagingServerImpl struct {
+	roomDb RoomDb
+}
+
+// Creates a room.
+func (s *messagingServerImpl) CreateRoom(ctx context.Context, in *pb.CreateRoomRequest) (*pb.Room, error) {
+	return s.roomDb.Create(in.GetRoom())
+}
```
pongad: Is there any reason `RoomDb` doesn't implement these methods directly?

server/room_db.go
```diff
@@ -0,0 +1,225 @@
...
+)
+
+func NewRoomDb() RoomDb {
+	return &roomDb{
+		uid:   &uniqID{},
+		token: NewTokenGenerator(),
+
+		mu:    sync.Mutex{},
+		keys:  map[string]int{},
+		rooms: []roomEntry{},
```
pongad: `mu` and `rooms` are essentially set to zero value. We can just elide the initialization.

server/room_db.go
```diff
@@ -0,0 +1,225 @@
...
+	Get(name string) (*pb.Room, error)
+}
+
+type roomEntry struct {
+	room    *pb.Room
+	deleted bool
+}
+
+type roomDb struct {
+	uid   *uniqID
```
pongad: I think this can just be the value, not the pointer, then you can also elide the init in `NewRoomDb`.

PR #47, opened 2018-11-20
Implement Messaging Blurb CRUD Methods

landrito:
Depends on #46


server/blurb_db.go
```diff
@@ -0,0 +1,214 @@
...
+	pb "github.com/googleapis/gapic-showcase/server/genproto"
+	"google.golang.org/genproto/protobuf/field_mask"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+func NewBlurbDb() BlurbDb {
+	return &blurbDb{
+		token:      NewTokenGenerator(),
+		blurbMu:    sync.Mutex{},
```
noahdietz: to go with pongad's comment in another PR, this is the zero value of a Mutex, and it can be omitted from struct initialization

landrito: Done.

PR #49, opened 2018-11-20
Implement Blurb Streaming methods

landrito:
Depends on #48 

cc @lukesneeringer for schema changes.


server/messaging_service.go
```diff
@@ -135,21 +139,130 @@ func (s *messagingServerImpl) SearchBlurbs(ctx context.Context, in *pb.SearchBlu
 // This returns a stream that emits the blurbs that are created for a
 // particular chat room or user profile.
 func (s *messagingServerImpl) StreamBlurbs(in *pb.StreamBlurbsRequest, stream pb.Messaging_StreamBlurbsServer) error {
+	parent := in.GetName()
+	if err := s.validateParent(parent); err != nil {
+		return err
+	}
+
+	expireTime := time.Unix(int64(0), int64(0))
+	expireTime, _ = ptypes.Timestamp(in.GetExpireTime())
```
noahdietz: why double set here? Is Line 147 to force a type?

landrito: Its actually to ensure that expiretime is the zero value in case there is an error. I made it more explicit.

PR #57, opened 2019-01-04
Add gapic-generator-go smoke test

landrito:


schema/identity.proto
```diff
@@ -31,6 +31,10 @@ option (google.api.metadata) = {
 
 // A simple identity service.
 service Identity {
+  // This service is meant to only run locally on the port 7469 (keypad digits
+  // for "show").
+  option (google.api.default_host) = "localhost:7469";
```
noahdietz: Just so you know, `gapic-generator-go` will append `:443` to whatever you put here and it defaults to a secure dial as it is meant to be used with remote hosts. So to use the gapic against a locally running server, you'd need to explicitly configure the address and connection dialing in the client initialization. You *cannot* use the default client against a local server.

noahdietz: lmk if you want to talk about this more. This was a problem for and feedback from use in Kiosk

landrito: Yeah I caught that in testing my CLI refactor. I added some code to the cli output to handle it. I think my case and kiosk is a bit specific so I wouldn't argue for changing it.

PR #59, opened 2019-01-08
Refactor CLI to use gcli generator

landrito:


cmd/gapic-showcase/connect.go
```diff
@@ -0,0 +1,105 @@
...
+		stream, err := MessagingClient.Connect(ctx)
+
+		out, err := os.OpenFile(ConnectOutFile, os.O_APPEND|os.O_WRONLY, os.ModeAppend)
+		if err != nil {
+			return err
+		}
+
+		// start background stream receive
+		go func() {
+			var res *genprotopb.Blurb
```
noahdietz: `messaging.proto` changed to `StreamBlurbsResponse`, looks like you need to regenerate the CLI for this PR

landrito: Done!

PR #60, opened 2019-01-09
Refactor testing.proto schema

landrito:
Deletes prior server code since I am going to refactor that.

I'll regenerate after this is merged.


schema/testing.proto
```diff
@@ -55,123 +92,13 @@ service Testing {
   //
   // In cases where a test involves registering a final answer at the
   // end of the test, this method provides the means to do so.
-  rpc RegisterTest(RegisterTestRequest) returns (google.protobuf.Empty) {
+  rpc CheckAnswer(CheckAnswerRequest) returns (CheckAnswerResponse) {
```
lukesneeringer: The noun here is "Test". `CheckTest`, perhaps?

schema/testing.proto
```diff
@@ -238,4 +230,128 @@ message Test {
...
   // A description of the test.
   string description = 3;
+
+  // A blueprint is an explicit definition of methods and requests that are needed
+  // to be made to test this specific test case. Ideally this would be represented
+  // by something more robust like CEL, but as of writing this, I am unsure if CEL
+  // is ready.
+  message Blueprint {
+    // The id of this blueprint.
+    string id = 1;
```
lukesneeringer: What are examples of this value? (I expect this should be called something like `code`, since ID is overloaded.)

schema/testing.proto
```diff
@@ -238,4 +230,128 @@ message Test {
...
+    // A description of this blueprint.
+    string description = 2;
+
+    // A message representing a method invocation.
+    message Invocation {
+      // The fully qualified name of the showcase method to be invoked.
+      string method = 1;
+
+      // The request to be made if a specific request is necessary.
+      google.protobuf.Any request = 2;
```
lukesneeringer: The `type_url` is not going to do anyone any good here; this could just be a bytes:

```proto
bytes serialized_request = 2;
```

schema/testing.proto
```diff
@@ -238,4 +230,128 @@ message Test {
...
+    }
+
+    // An ordered list of method calls that can be called to trigger this test.
+    repeated Invocation invocations = 3;
```
lukesneeringer: This model will not stand up to client-side streaming since it only provides for one request body.

That might be fine though? Having a main request and a `repeated additional_requests` or something similar would actually be much _easier_ than making the request repeated...and you can always add the additional component later if/when you need it.

schema/testing.proto
```diff
@@ -238,4 +230,128 @@ message Test {
...
+message Issue {
+  // The different potential types of issues.
+  enum Type {
+    TYPE_UNSPECIFIED = 0;
+
+    // The test was never instrumented.
+    SKIPPED = 1;
+
+    // The test was started but never confirmed.
+    PENDING = 2;
```
lukesneeringer: How does Showcase know this one?

schema/testing.proto
```diff
@@ -238,4 +230,128 @@ message Test {
...
+  // The test to be deleted.
+  string name = 1 [(google.api.resource_reference) = "Test"];
+}
+
+message CheckAnswerRequest {
+  // The test to have an answer registered to it.
+  string name = 1 [(google.api.resource_reference) = "Test"];
+
+  // The answer from the test.
+  google.protobuf.Any answer = 2;
```
lukesneeringer: Are you able to know what type should be sent back? If so, using a `bytes` here is likely to be better.

schema/testing.proto
landrito: Agreed.

schema/testing.proto
landrito: Added a resource path and renamed to `name`.

schema/testing.proto
landrito: Ah agreed. Changed.

schema/testing.proto
landrito: Accepted.

schema/testing.proto
landrito: If the initial request in any of the blueprints was called then the test is pending, unless it has been confirmed or failed.

schema/testing.proto
landrito: Done!

landrito:
@lukesneeringer this is ready for another look!

schema/testing.proto
```diff
@@ -55,123 +91,13 @@ service Testing {
   //
   // In cases where a test involves registering a final answer at the
   // end of the test, this method provides the means to do so.
-  rpc RegisterTest(RegisterTestRequest) returns (google.protobuf.Empty) {
+  rpc CheckTest(CheckTestRequest) returns (CheckTestResponse) {
```
lukesneeringer: Preemptive apologies: I know `CheckTest` was my suggestion, but it feels a bit vague now. Perhaps `VerifyTest` would be better?

schema/testing.proto
```diff
@@ -180,7 +106,7 @@ message RegisterTestRequest {
 message Session {
   // The name of the session. The ID must conform to ^[a-z]+$
   // If this is not provided, Showcase chooses one at random.
-  string name = 1 [(google.api.resource).path = "sessions/*"];
+  string name = 1 [(google.api.resource).path = "sessions/{session}"];
 
   // The specification versions understood by Showcase.
   enum Version {
```
lukesneeringer: Document what the user gets if sending `VERSION_UNSPECIFIED` (I assume it will be the latest.)

PR #61, opened 2019-01-09
Only track coverage for user written files.

landrito:


.circleci/config.yml
```diff
@@ -60,7 +60,9 @@ jobs:
           command: go get -v -d -t ./cmd/gapic-showcase && go install ./cmd/gapic-showcase
       - run:
           name: Run tests
-          command: go test ./... -coverprofile=coverage.txt -covermode=atomic
+          command: |
+            go test ./server -coverprofile=coverage.txt -covermode=atomic
+            go test ./client # Don't run coverage for generated tests.
```
noahdietz: you could also use [Go build tags](https://dave.cheney.net/2014/09/28/using-build-to-switch-between-debug-and-release) to separate test suites. Just another way of doing it!

landrito: Interesting. Good to know!

PR #64, opened 2019-01-29
Expose Messaging and Identity service. And fix missing cov

landrito:


server/messaging_service.go
```diff
@@ -484,7 +484,7 @@ func (s *messagingServerImpl) StreamBlurbs(in *pb.StreamBlurbsRequest, stream pb
 
 	expireTime, err := ptypes.Timestamp(in.GetExpireTime())
 	if err != nil {
-		expireTime = time.Unix(int64(0), int64(0))
+		return status.Error(codes.InvalidArgument, err.Error())
```
noahdietz: is it possible that `in.GetExpireTime()` returns `nil`? If so, what would be the behavior of `ptypes.Timestamp()`? Would it emit an error?

landrito: I think it treats it like UTC 0,0 since protobuf has no field presence.
https://github.com/golang/protobuf/blob/master/ptypes/timestamp.go#L95

PR #65, opened 2019-01-29
v0.0.10 Release Candidate.

landrito:
Relies on #64

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,6 @@
+# Release History
+
+### 0.0.10 / 2019-01-29
+- Expose messaging and identity services when running `gapic-showcase run`.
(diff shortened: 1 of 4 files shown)


cmd/gapic-showcase/run.go
```diff
@@ -53,8 +53,15 @@ func init() {
 			}
 			s := grpc.NewServer(opts...)
 			defer s.GracefulStop()
+
+			// Register Services to the server.
```
noahdietz: do you really want this commit in this PR? Shouldn't this branch contain only the version changes?

landrito: I'll rebase after #64 is merged and this will disappear from this PR.

PR #67, opened 2019-02-06
Observer pattern for testing infra

landrito:


server/observer.go
```diff
@@ -0,0 +1,199 @@
...
+
+type showcaseStream struct {
+	info     *grpc.StreamServerInfo
+	registry *showcaseObserverRegistry
+
+	grpc.ServerStream
+}
+
+func (s *showcaseStream) SendMsg(m interface{}) error {
+	s.registry.mu.Lock()
```
noahdietz: Is the point of locking the stream message transfer with the same mutex as Observer mgmt to prevent the observer from being removed mid stream? If thats not the purpose, maybe we should have separate mutexes for stream transfer & observer mgmt.

server/observer.go
```diff
@@ -0,0 +1,199 @@
...
+
+func (r *showcaseObserverRegistry) UnaryInterceptor(
+	ctx context.Context,
+	req interface{},
+	info *grpc.UnaryServerInfo,
+	handler grpc.UnaryHandler) (interface{}, error) {
+	resp, err := handler(ctx, req)
+
+	if info.FullMethod == "/google.showcase.v1alpha3.Testing/DeleteTest" {
+		err = r.deleteTestHandler(req)
```
noahdietz: Out of curiosity, is it possible that a delete call beats a unary/stream call using the same observer, grabs the mutex, and deletes the observer? Or are the tests structured such that the unary/stream calls are run in sequence and a delete call will only be complete once all of the related unary tests are complete?

server/observer.go
```diff
@@ -0,0 +1,199 @@
...
+	"errors"
+	"fmt"
+	"sync"
+
+	pb "github.com/googleapis/gapic-showcase/server/genproto"
+	"google.golang.org/grpc"
+)
+
+// UnaryObserver provides an interface for observing unary requests and responses.
+type UnaryObserver interface {
```
pongad: I think I know what you're trying to do here, but I think it is confusingly similar to UnaryServerInterceptor.

My current opinion, would it be better (or more flexible) to create a helper like this?
```
func UnaryServerChain(... grpc.UnaryServerInterceptor) grpc.UnaryServerInterceptor
```
Then it's up to the users to work on things themselves. Instead of adding/deleting interceptors, they can arrange for interceptors to enable/disable themselves as well.

landrito: Agreed but chaining `UnaryServerInterceptors` is not possible.

It is explicitely stated in the grpc documentation: https://godoc.org/google.golang.org/grpc#UnaryInterceptor.

The reason why chaining is not possible is because the handler passed in to the interceptor should only be called once since the handler completes the normal execution of a unary RPC.

server/observer.go
landrito: Correct, it is to prevent the observer from being removed mid stream.

server/observer.go
landrito: Since all the observers are behind the same lock and all access to the observers will lock the mutex. Look down at the impl of the deleteTestHandler, it locks if it is going to attempt to delete the test. Maybe I'm missing something?

noahdietz: nvm i talked myself out of confusion as I drafted a response. Thanks.

server/observer.go
pongad: I think you misunderstood me. I meant, we can wrap a list of user-provided interceptors, do the chaining logic ourselves, like you did it here. Then present grpc with only on interceptor. Am I making sense? I believe that should work right?

landrito: Yeah I think that strategy could work. We'd have to mock the handler to ensure that the actual handler only gets called once.

```go

func UnaryServerChain(interceptors ...grpc.UnaryServerInterceptor) grpc.UnaryServerInterceptor {
	return func(ctx context.Context, req interface{}, info *UnaryServerInfo, handler UnaryHandler) (interface{}, error) {
		resp, err := handler(ctx, req)
		mockHandler := func(context.Context, interface{}) (interface{}, error) { return resp, err }
		for _, interceptor := range interceptors {
			interceptor(ctx, req, info, mockHandler)
		}
		return resp, err
	}
}
```

For UnaryInterceptors I think this makes sense. But to do something similar for stream interceptors will cause a lot of extra code down the line. Every test case will need to implement a `grpc.ServerStream` to hook into the streaming request and responses. 

I'd argue to keep doing the stream observations this way, to keep that interface simple and then to keep the observers this way to have them be consistent.

pongad: I think that sounds reasonable :P

landrito: Woops I had a typo, 

Keep stream observations this way and keep the unary observations this way to keep them consistent.

server/observer.go
```diff
@@ -0,0 +1,199 @@
+// Copyright 2018 Google LLC
```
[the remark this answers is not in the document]
landrito: Fixed!

PR #68, opened 2019-02-07
Testing Sessions and Test Case Interface

landrito:
This PR includes implementation for the Test and Session resources.

### DIFF
--- a/server/session.go
+++ b/server/session.go
@@ -119,12 +119,13 @@ func (s *sessionImpl) GetReport() *pb.ReportSessionResponse {
 
 func (s *sessionImpl) ListTests(in *pb.ListTestsRequest) (*pb.ListTestsResponse, error) {
 	start, err := s.token.GetIndex(in.GetPageToken())
+	if err != nil {
+		return nil, err
+	}
+
 	pageSize := in.GetPageSize()
 	if pageSize == 0 {
 		pageSize = 10
-	}
-	if err != nil {
-		return nil, err
 	}
 
 	offset := 0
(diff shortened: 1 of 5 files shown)


server/session.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+}
+
+func (s *sessionImpl) ListTests(in *pb.ListTestsRequest) (*pb.ListTestsResponse, error) {
+	start, err := s.token.GetIndex(in.GetPageToken())
+	pageSize := in.GetPageSize()
+	if pageSize == 0 {
+		pageSize = 10
+	}
+	if err != nil {
```
noahdietz: Nit: the error check should follow the function that emits it

server/session.go
```diff
@@ -0,0 +1,194 @@
...
+		codes.NotFound,
+		"A test with name %s not found.", name)
+}
+
+func (s *sessionImpl) RegisterTests(tests []Test) {
+	for _, test := range tests {
+		i := len(s.tests)
+		s.tests = append(s.tests, testEntry{test: test})
+		s.keys[test.GetName()] = i
+		if obs, ok := test.(UnaryObserver); ok {
```
noahdietz: Same here. A type switch might be be better

landrito: I don't think a type switch will be applicable here since the type switch will only run the first type that it matches and fallthroughs in type switches dont work.

https://play.golang.org/p/gSlDbQGzVqX

server/session.go
landrito: Done.

server/session.go
noahdietz: Ah I missed the fact that `A Test will also implement at least one of` the observers. Cool

PR #78, opened 2019-02-21
fixes for github_release job

noahdietz:
Fixes #78 

* switches to use of `CIRCLE_TAG` instead of installing & running the `gapic-showcase` CLI to determine the version, because the affected jobs only run on tag pushes
* hard code github org & repo name


noahdietz:
The changes were tested on a throw-away tag and it worked: https://circleci.com/gh/googleapis/gapic-showcase/999

I will clean up.

PR #88, opened 2019-03-05
Utility Commands Cleanup

landrito:
This PR cleans up the functionality of the utils folder by doing the following:

- Consolidates the generation of the CLI, Client, and generated protos into util/cmd/compile_protos.go
- Moves the publishing of github resources and pushing the docker image from the circle config to the release script.
- Makes all util files no longer rely on the GOPATH env variable


noahdietz:
LGTM but to be clear, this removes the automation of releases entirely? Someone needs to run the release script now?

landrito:
Correct. This is essentially the status quo since the releasing to github and gcr.io through circle would require us to store github tokens and gcr tokens on circle, which I think is bad practice.

landrito:
Essentially the release flow would be:

1. go run util/cmd/bump_version/main.go --patch
1. PR with changelog
1. go run util/cmd/release/main.go
1. update release notes with changelog (ideally the prior step would grab the latest info from the changelog)

landrito:
Sorry to be more clear, Automated releasing through circle has never worked since someone on the team pointed out that storing github tokens on circle was a bad idea.

landrito:
Redoing this in separate PRs.

PR #89, opened 2019-03-06
Testing Service Implementation

landrito:


server/testing_service.go
```diff
@@ -0,0 +1,184 @@
...
+	pb "github.com/googleapis/gapic-showcase/server/genproto"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// NewTestingServer returns a new TestingServer for the Showcase API.
+func NewTestingServer(observerRegistry GrpcObserverRegistry) pb.TestingServer {
+	name := fmt.Sprintf("sessions/-")
+	defaultSession := NewSession(name, pb.Session_V1_LATEST, observerRegistry)
+	sessions := []sessionEntry{sessionEntry{session: defaultSession}}
```
noahdietz: Nit: You can omit the struct field name when in-lining new structs & values. Especially when we are including only 1 of 2 fields (and here it seems a little repetitive).

I'm not 100% sure on the convention, but this is just a small nit.

server/testing_service.go
```diff
@@ -0,0 +1,184 @@
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	seshProto := req.GetSession()
+	id := s.uid.next()
+	name := fmt.Sprintf("sessions/%d", id)
+	sesh := NewSession(name, seshProto.GetVersion(), s.observerRegistry)
+
+	index := len(s.sessions)
+	s.sessions = append(s.sessions, sessionEntry{session: sesh})
```
noahdietz: Same struct in-lining comment here

server/testing_service.go
```diff
@@ -0,0 +1,184 @@
...
+
+func (s *testingServerImpl) ListSessions(_ context.Context, in *pb.ListSessionsRequest) (*pb.ListSessionsResponse, error) {
+	start, err := s.token.GetIndex(in.GetPageToken())
+	if err != nil {
+		return nil, err
+	}
+
+	offset := 0
+	sessions := []*pb.Session{}
+	for _, entry := range s.sessions[start:] {
```
noahdietz: I don't remember the `token.GetIndex()` logic at all, but would it be possible for me to supply a page token that resulted in a `start` index that's out of bounds here? It might be a concern if a gapic is jacking up the request payload, but maybe its not a real problem

server/testing_service.go
```diff
@@ -0,0 +1,184 @@
...
+			continue
+		}
+		sessions = append(sessions, SessionProto(entry.session))
+		if len(sessions) >= int(in.GetPageSize()) {
+			break
+		}
+	}
+
+	nextToken := ""
+	if start+offset < len(s.sessions) {
```
noahdietz: Super nit: could be
```
if next := start + offset; next < len(s.sessions) {
  nextToken = s.token.ForIndex(next)
}
```

server/testing_service.go
```diff
@@ -0,0 +1,184 @@
...
+
+	entry := s.sessions[i]
+	s.sessions[i] = sessionEntry{session: entry.session, deleted: true}
+
+	return &empty.Empty{}, nil
+}
+
+func (s *testingServerImpl) ReportSession(_ context.Context, req *pb.ReportSessionRequest) (*pb.ReportSessionResponse, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
```
noahdietz: Super nit: new line after this

server/testing_service.go
```diff
@@ -0,0 +1,184 @@
...
+			return SessionProto(entry.session), nil
+		}
+	}
+
+	return nil, status.Errorf(
+		codes.NotFound, "A session with name %s not found.",
+		name)
+}
+
+func (s *testingServerImpl) ListSessions(_ context.Context, in *pb.ListSessionsRequest) (*pb.ListSessionsResponse, error) {
```
noahdietz: `ListTests` below uses a lock and so does the above `GetSession`. Should we have one here as well?

server/testing_service.go
```diff
@@ -0,0 +1,184 @@
...
+	entry := s.sessions[i]
+	s.sessions[i] = sessionEntry{session: entry.session, deleted: true}
+
+	return &empty.Empty{}, nil
+}
+
+func (s *testingServerImpl) ReportSession(_ context.Context, req *pb.ReportSessionRequest) (*pb.ReportSessionResponse, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if i, ok := s.keys[req.GetName()]; ok {
```
noahdietz: `ok && !s.sessions[i].deleted {` ?

server/testing_service.go
```diff
@@ -0,0 +1,184 @@
...
+
+	return SessionProto(sesh), nil
+}
+
+func (s *testingServerImpl) GetSession(_ context.Context, req *pb.GetSessionRequest) (*pb.Session, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	name := req.GetName()
+	if i, ok := s.keys[name]; ok {
```
noahdietz:
```go
if i, ok := s.keys[name]; ok && !s.sessions[i].deleted {
  return SessionProto(s.sessions[i].session), nil
}
```

server/testing_service.go
landrito: It's definitely possible. I tried to deter by making it hard to provide a valid page token that will result in a start that is out of bounds. This logic is everywhere so I'll make an issue to add a guard to protect from that.

server/testing_service.go
landrito: Resolving due to offline conversation.

server/testing_service.go
landrito: Resolving due to offline conversation.

server/testing_service.go
landrito: Done.

server/testing_service.go
landrito: Done.

server/testing_service.go
landrito: Ah good catch thanks! Done.

server/testing_service.go
landrito: Done.

server/testing_service.go
landrito: Done.

server/testing_service.go
landrito: I filed: https://github.com/googleapis/gapic-showcase/issues/90

PR #93, opened 2019-03-13
Server dir cleanup

landrito:
- Moves service implementations into their own separate directory.
  - I am trying to place the v1 test spec into a separate directory. The test cases have dependencies on some things in the server directory and would cause a circular dependency when the testing service depends on the v1 spec.
- Adds some documentation to exported things. 
- Addresses some VSCode suggestions


server/services/common_test.go
```diff
@@ -0,0 +1,31 @@
+// Copyright 2019 Google LLC
```
noahdietz: `_test.go` is a suffix reserved for files that contain test functions. We might consider naming this something like `test_utils.go` or `test_common.go`.

landrito: Done!

PR #95, opened 2019-03-22
Fix path templates to start with /

alexander-fenster:
gapic-generator fails to generate clients for Showcase v0.0.13 because of these path templates that don't start with `/`.

```
ERROR: google/showcase/v1alpha3/messaging.proto:93:3: http: In path template 'v1alpha3/{parent=users/*/profile}/blurbs': effective path must start with leading '/'.
```

Fixing that (and will release v0.0.14 when fixed).

### DIFF
--- a/schema/messaging.proto
+++ b/schema/messaging.proto
@@ -95,7 +95,7 @@ service Messaging {
       post: "/v1alpha3/{parent=rooms/*}/blurbs"
       body: "*"
       additional_bindings: {
-        post: "v1alpha3/{parent=users/*/profile}/blurbs"
+        post: "/v1alpha3/{parent=users/*/profile}/blurbs"
         body: "*"
       }
     };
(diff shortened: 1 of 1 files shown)


noahdietz:
The build fails building the `gapic-showcase` CLI because of an issue with the `server` impl that is a dependency of `cmd/gapic-showcase/run.go`.
```
go get -v -d -t ./cmd/gapic-showcase && go install ./cmd/gapic-showcase
...
# github.com/googleapis/gapic-showcase/server/services
server/services/operations_service.go:34:67: cannot use operationsServerImpl literal (type *operationsServerImpl) as type longrunning.OperationsServer in return argument:
	*operationsServerImpl does not implement longrunning.OperationsServer (missing WaitOperation method)
Exited with code 2
```

Looks like the interface was not fully implemented @landrito ?

PR #106, opened 2019-03-26
Move protos into their proper proto path

landrito:
This makes it easier to invoke `protoc` on the showcase protos by making it such that the schema directory can be passed as a `--proto_path` to `protoc`. 

```sh
protoc --proto_path=gapic-showcase/schema \
  --proto_path=api-common-protos \
  --generator_out=. \
  gapic-showcase/schema/**/*.proto
```


landrito:
This is not ready for review. I just realized I needed to update the scripts.

landrito:
Okay this is ready for review!

PR #109, opened 2019-03-27
Update release script

landrito:
- Adds documentation
- Remove dependency on $GOPATH
- Only includes the dependencies in the proto


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "api-common-protos"]
```
noahdietz: Should this be in here if its in the other CL?

landrito: It will get handled when I merge from master.

PR #111, opened 2019-04-01
 Add CI to regenerate client, cmd, and server source if needed

landrito:
- Added a job regenerate to run on master which does the following
  - Checks out the repo
  - Calls compile_protos
  - Check if theres a diff
  - If theres a diff push to master

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,6 +18,8 @@ workflows:
       - regeneration-check:
           filters: *all_commits
       - regenerate:
+          requires:
+            - regeneration-check
           filters:
             branches:
               only: master
@@ -186,33 +188,18 @@ jobs:
           name: Print instructions if failed.
           command: echo "Regenerating generated source will cause the build to break, please run 'go run ./util/compile_protos' and fix the build."
           when: on_fail
+      - persist_to_workspace:
+          root: /go
+          paths:
+            - src/src/github.com/googleapis/gapic-showcase
 
   regenerate:
     docker:
-      - image: golang:1.12
+      - image: alpine.git:latest
     working_directory: /go/src/github.com/googleapis/gapic-showcase
     steps:
-      - checkout
-      - run:
-          name: Get proto dependencies
-          command: git submodule init && git submodule update
-      - run:
-          name: Install protoc
-          command: |
-            apt-get update && apt-get install -y unzip
-            curl -o ~/protoc3.zip -L https://github.com/google/protobuf/releases/download/v3.6.1/protoc-3.6.1-linux-x86_64.zip
-            unzip ~/protoc3.zip -d ~/protoc3
-            mv ~/protoc3/bin/* /usr/local/bin/
-            mv ~/protoc3/include/* /usr/local/include/
-      - run:
-          name: Get generators
-          command: |
-              go get github.com/golang/protobuf/protoc-gen-go
-              go get github.com/googleapis/gapic-generator-go/cmd/protoc-gen-go_cli
-              go get github.com/googleapis/gapic-generator-go/cmd/protoc-gen-go_gapic
-      - run:
-          name: Compile protos
-          command: go run ./util/cmd/compile_protos
+      - attach_workspace:
+          at: /go/src/src/github.com/googleapis/gapic-showcase
       - run:
           name: Check if compilation caused diff.
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -182,7 +182,6 @@ jobs:
             # Exit early if no diff was found
             [[ ! -z $DIFF_FOUND ]] || exit 0
             go test ./server/...
-            go test ./client
       - run:
           name: Print instructions if failed.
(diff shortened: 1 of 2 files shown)


landrito:
Hold off on reviewing this. I'm going to retry the approach.

.circleci/config.yml
```diff
@@ -15,6 +15,16 @@ workflows:
           filters: *all_commits
       - protobufjs-load-test:
           filters: *all_commits
+      - regenerate:
+          requires:
+            - build
```
noahdietz: I feel like these should be run after `regenerate`. It's already happened where we've generated the CLI and it was broken. I'm not sure that running all of these tests prior to regeneration everything would help a lot. Unless I'm missing something?

landrito: Agreed. Removed those requires.

landrito:
For context, this solves two problems:

1. The need to keep generated source up to date.
1. There being slight differences between generated messages and services across protoc versions and platforms

I originally wanted this to just be a PR check that ran the compile_protos and failed if there was a diff. I found there were slight errors between protoc version and platforms that ran the generation. To keep things consistent the master branch will simply use the generated clients from the circle ci job.

I need to do some thinking about how to handle a PR which will push an update to the schema that will break the implementation. I am probably gonna add a PR check that simply regenerates and runs the tests.

landrito:
I added that PR check. PTAL.

.circleci/config.yml
```diff
@@ -139,6 +147,78 @@ jobs:
...
+          name: Get generators
+          command: |
+              go get github.com/golang/protobuf/protoc-gen-go
+              go get github.com/googleapis/gapic-generator-go/cmd/protoc-gen-go_cli
+              go get github.com/googleapis/gapic-generator-go/cmd/protoc-gen-go_gapic
+      - run:
+          name: Compile protos
+          command: go run ./util/cmd/compile_protos
+      - run:
+          name: Check if compilation caused diff.
```
noahdietz: So this PR would trigger a diff from its regeneration. Is that expected? There aren't any proto changes in this PR (but maybe you were saving the regeneration from some earlier changes to test this?)

landrito: Yeah I think there were some earlier changes that I forgot to run the script for (good thing I'm automating it 😄)

landrito:
I made the tests depend on the regeneration and the push to github depend on the tests! This is ready for another look! PTAL.

landrito:
Made the push job push to staging to test the job. Will update that step to push to master after we confirm it works.

.circleci/config.yml
```diff
@@ -38,14 +48,50 @@ workflows:
...
 jobs:
   build:
+    docker:
+      - image: golang:1.12
+    working_directory: /go/src/github.com/googleapis/gapic-showcase
+    steps:
+      - checkout
+      - run:
+          name: Get proto dependencies
+          command: git submodule init && git submodule update
```
noahdietz: curious, will we remove the submodule after all the necessary annotations are merged into master? Or will this persist?

landrito: Merged into master of googleapis/googleapis? I think keeping the submodule makes sense since it shows where the dependencies of the schemas come from. I'll probably move the submodule into schema though to make that even more apparent.

noahdietz: api-common-protos master, but yeah. 

Yeah moving it to the schema directory seems like a good idea imo.

landrito: Moving to schema in #112.

.circleci/config.yml
```diff
@@ -74,6 +120,50 @@ jobs:
...
+            git config user.name "googleapis-publisher"
+            # Push changes to branch.
+            export BRANCH="build-$(date +%Y%m%d%H%M%S)"
+            git checkout -b ${BRANCH}
+            git add .
+            git commit -m "Regenerate cmd, client, and server directories."
+            git push --quiet https://${GITHUB_TOKEN}@github.com/googleapis/gapic-showcase.git ${BRANCH}
+            # Create a PR with hub.
+            go get github.com/github/hub
+            hub pull-request \
```
landrito: FYI this will use the the env var GITHUB_TOKEN to make this PR based on this documentation: https://hub.github.com/hub.1.html#github-oauth-authentication

noahdietz: this is super tight

.circleci/config.yml
```diff
@@ -74,6 +120,50 @@ jobs:
...
+      GO111MODULE: "on"
+    working_directory: /go/src/github.com/googleapis/gapic-showcase
+    steps:
+      - attach_workspace:
+          at: /go
+      - run:
+          name: Check if we are using generated source.
+          command: |
+            git status --porcelain
+            git diff --exit-code --quiet || echo 'export DIFF_FOUND="true"' >> $BASH_ENV
```
noahdietz: If it's not a lot of trouble, can we make the diff checking more pointed? Just diff the protos, generated pkgs, etc.

I don't want to close a bunch of PRs & branches because we are updating docs or utility files.

landrito: Done!

PR #127, opened 2019-04-17
feat: Node.js gRPC server

alexander-fenster:
This PR adds a Node.js gRPC server that can serve RPCs defined in `echo.proto` (including long running operations). A trivial baseline test checks the basic functionality of the server.

Usage:

Start server:
```sh
$ npm install
$ node build/src/index.js --verbose    # will start on port 7469
```

Play with client:
```sh
$ docker run --rm --network host gcr.io/gapic-images/gapic-showcase:0.1.1 \
  echo --address host.docker.internal:7469 echo --response content --response.content okay

$ docker run --rm --network host gcr.io/gapic-images/gapic-showcase:0.1.1 \
  echo --address host.docker.internal:7469 wait --end ttl --end.ttl.seconds 5 \
  --follow --response success --response.success.content okay
```

@JustinBeckwith This will be used in GAX testing. Asking you to review TypeScript stuff :)

Thanks!

### DIFF
--- a/nodejs-server/package.json
+++ b/nodejs-server/package.json
@@ -1,0 +1,39 @@
+{
+  "name": "showcase-nodejs-echo-server",
+  "version": "0.0.1",
+  "description": "",
+  "main": "build/src/index.js",
+  "types": "build/src/index.d.ts",
+  "files": [
+    "build/src"
+  ],
+  "license": "Apache-2.0",
+  "keywords": [],
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "check": "gts check",
+    "clean": "gts clean",
+    "precompile": "mkdir -p pbjs-genfiles && pbjs -t static-module -o pbjs-genfiles/echo.js -p ../schema -p node_modules/google-proto-files google/showcase/v1beta1/echo.proto && pbts -o pbjs-genfiles/echo.d.ts pbjs-genfiles/echo.js",
+    "compile": "mkdir -p build && cp -r pbjs-genfiles build/ && tsc -p .",
+    "fix": "gts fix",
+    "prepare": "npm run compile",
+    "pretest": "npm run compile",
+    "posttest": "npm run check",
+    "start": "node build/src/index.js"
+  },
+  "devDependencies": {
+    "@types/long": "^4.0.0",
+    "@types/node": "^11.13.4",
+    "@types/yargs": "^13.0.0",
+    "gts": "^0.9.0",
+    "protobufjs": "^6.8.8",
+    "typescript": "~3.1.0"
+  },
+  "dependencies": {
+    "@grpc/proto-loader": "^0.5.0",
+    "google-proto-files": "^0.20.0",
+    "grpc": "^1.19.0",
+    "long": "^4.0.0",
+    "yargs": "^13.2.2"
+  }
+}
--- a/nodejs-server/package.json
+++ b/nodejs-server/package.json
@@ -1,7 +1,7 @@
 {
   "name": "showcase-nodejs-echo-server",
   "version": "0.0.1",
-  "description": "",
+  "description": "gRPC server implementing gapic-showcase Echo service",
   "main": "build/src/index.js",
   "types": "build/src/index.d.ts",
   "files": [
@@ -10,30 +10,30 @@
   "license": "Apache-2.0",
   "keywords": [],
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1",
-    "check": "gts check",
+    "test": "sh test/test.sh",
+    "lint": "gts check",
     "clean": "gts clean",
     "precompile": "mkdir -p pbjs-genfiles && pbjs -t static-module -o pbjs-genfiles/echo.js -p ../schema -p node_modules/google-proto-files google/showcase/v1beta1/echo.proto && pbts -o pbjs-genfiles/echo.d.ts pbjs-genfiles/echo.js",
     "compile": "mkdir -p build && cp -r pbjs-genfiles build/ && tsc -p .",
     "fix": "gts fix",
     "prepare": "npm run compile",
     "pretest": "npm run compile",
-    "posttest": "npm run check",
+    "posttest": "npm run lint",
     "start": "node build/src/index.js"
   },
   "devDependencies": {
     "@types/long": "^4.0.0",
     "@types/node": "^11.13.4",
-    "@types/yargs": "^13.0.0",
-    "gts": "^0.9.0",
+    "@types/yargs-parser": "^13.0.0",
+    "gts": "1.0.0-0",
     "protobufjs": "^6.8.8",
-    "typescript": "~3.1.0"
+    "typescript": "~3.4.3"
   },
   "dependencies": {
     "@grpc/proto-loader": "^0.5.0",
     "google-proto-files": "^0.20.0",
     "grpc": "^1.19.0",
     "long": "^4.0.0",
-    "yargs": "^13.2.2"
+    "yargs-parser": "^13.0.0"
   }
 }
--- a/nodejs-server/package.json
+++ b/nodejs-server/package.json
@@ -1,7 +1,7 @@
 {
-  "name": "showcase-nodejs-echo-server",
+  "name": "gapic-showcase-server",
   "version": "0.0.1",
-  "description": "gRPC server implementing gapic-showcase Echo service",
+  "description": "gRPC server implementing gapic-showcase service",
   "main": "build/src/index.js",
   "types": "build/src/index.d.ts",
   "files": [
--- a/nodejs-server/src/echoServer.ts
+++ b/nodejs-server/src/echoServer.ts
@@ -1,0 +1,316 @@
+import * as grpc from 'grpc';
+import * as Long from 'long';
+
+import * as echoTypes from '../pbjs-genfiles/echo.js';
+
+import showcaseV1Beta1 = echoTypes.google.showcase.v1beta1;
+import google = echoTypes.google;
+import longrunning = google.longrunning;
+import {OperationsServer} from './operationsServer.js';
+
+/**
+ * Implements Echo server based on echo.proto.
+ */
+export class EchoServer {
+  private requestCount: number;
+  private paginationOutput: {[index: string]: string[]};
+  private paginationRequest: {[index: string]: string|null|undefined};
+  private operationsServer: OperationsServer;
+  private verbose: boolean;
+
+  constructor(verbose: boolean, operationsServer: OperationsServer) {
+    this.requestCount = 0;
+    this.paginationOutput = {};
+    this.paginationRequest = {};
+    this.operationsServer = operationsServer;
+    this.verbose = verbose;
+  }
+
+  private log(request: number, ...args: Array<string|{}>) {
+    if (this.verbose) {
+      console.log(`[EchoServer request #${request}]`, ...args);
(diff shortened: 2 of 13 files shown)


alexander-fenster:
cc: @bcoe FYI :)

nodejs-server/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "showcase-nodejs-echo-server",
+  "version": "0.0.1",
+  "description": "",
```
JustinBeckwith: It's worth adding something here :)

nodejs-server/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "main": "build/src/index.js",
+  "types": "build/src/index.d.ts",
+  "files": [
+    "build/src"
+  ],
+  "license": "Apache-2.0",
+  "keywords": [],
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "check": "gts check",
```
JustinBeckwith: it's a total nit, but we've been using `lint` here since it's a little more community standard

nodejs-server/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "posttest": "npm run check",
+    "start": "node build/src/index.js"
+  },
+  "devDependencies": {
+    "@types/long": "^4.0.0",
+    "@types/node": "^11.13.4",
+    "@types/yargs": "^13.0.0",
+    "gts": "^0.9.0",
+    "protobufjs": "^6.8.8",
+    "typescript": "~3.1.0"
```
JustinBeckwith: That's a sorta old version of TypeScript

nodejs-server/package.json
```diff
@@ -0,0 +1,39 @@
...
+  },
+  "dependencies": {
+    "@grpc/proto-loader": "^0.5.0",
+    "google-proto-files": "^0.20.0",
+    "grpc": "^1.19.0",
```
JustinBeckwith: grpc-js?

nodejs-server/src/echoServer.ts
```diff
@@ -0,0 +1,316 @@
...
+import google = echoTypes.google;
+import longrunning = google.longrunning;
+import {OperationsServer} from './operationsServer.js';
+
+/**
+ * Implements Echo server based on echo.proto.
+ */
+export class EchoServer {
+  private requestCount: number;
+  private paginationOutput: {[index: string]: string[]};
```
JustinBeckwith: For both of these - why not just use a Map?

nodejs-server/src/echoServer.ts
```diff
@@ -0,0 +1,316 @@
...
+    const status = EchoServer.requestToStatus(request);
+    const error = new Error(status.message) as grpc.ServiceError;
+    error.code = status.code;
+    return error;
+  }
+
+  /**
+   * Unary call example. Echoes the request.
+   */
+  echo(
```
JustinBeckwith: Just style, but use async instead of callback based

nodejs-server/src/echoServer.ts
```diff
@@ -0,0 +1,316 @@
...
+  /**
+   * Client streaming call example. Joins the list of words received from a
+   * stream into a string.
+   */
+  collect(
+      call: grpc.ServerReadableStream<showcaseV1Beta1.EchoRequest>,
+      callback: grpc.requestCallback<showcaseV1Beta1.EchoResponse>): void {
+    const requestId = ++this.requestCount;
+    const results: string[] = [];
+    let error: grpc.ServiceError|undefined = undefined;
```
JustinBeckwith: no need to assign to undefined, that's its natural state :)

nodejs-server/src/echoServer.ts
```diff
@@ -0,0 +1,316 @@
...
+    call.end();
+  }
+
+  /**
+   * Client streaming call example. Joins the list of words received from a
+   * stream into a string.
+   */
+  collect(
+      call: grpc.ServerReadableStream<showcaseV1Beta1.EchoRequest>,
+      callback: grpc.requestCallback<showcaseV1Beta1.EchoResponse>): void {
```
JustinBeckwith: same as other feedback, use async/await

nodejs-server/src/index.ts
```diff
@@ -0,0 +1,66 @@
+import * as protoLoader from '@grpc/proto-loader';
+import * as grpc from 'grpc';
+import * as path from 'path';
+import * as yargs from 'yargs';
```
JustinBeckwith: If all you're using here is the argument parser, I think this ships in it's own smaller package cc @bcoe

nodejs-server/src/operationsServer.ts
```diff
@@ -0,0 +1,172 @@
...
+      return;
+    }
+    this.log(requestId, 'getOperation waiting...');
+    operation.setCallback(() => {
+      const response = operation.getOperation();
+      this.log(requestId, 'getOperation response:', response);
+      callback(null, response);
+    });
+  }
+}
```
JustinBeckwith: newwwwwline

nodejs-server/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "prepare": "npm run compile",
+    "pretest": "npm run compile",
+    "posttest": "npm run check",
+    "start": "node build/src/index.js"
+  },
+  "devDependencies": {
+    "@types/long": "^4.0.0",
+    "@types/node": "^11.13.4",
+    "@types/yargs": "^13.0.0",
+    "gts": "^0.9.0",
```
JustinBeckwith: it may be worth trying this with the https://www.npmjs.com/package/gts/v/1.0.0-0 tag - it is much better, and will be the default v soon

nodejs-server/package.json
alexander-fenster: It does not have server support yet (AFAIK)

nodejs-server/src/echoServer.ts
alexander-fenster: I think `(call, callback) => void` is the interface that gRPC expects. This `EchoServer` class is being fed to `grpc.Server.addService()` and they expect certain API.

nodejs-server/src/index.ts
bcoe: :wave: hey, just use https://www.npmjs.com/package/yargs-parser in these cases; it's the parser used by yargs, without the CLI output bits.

nodejs-server/src/operationsServer.ts
bcoe: we should totes make our linter catch this, and we both would have saved typing in messages :stuck_out_tongue_closed_eyes:

nodejs-server/package.json
alexander-fenster: Changed to `lint`. I actually used `npx gts init` so unless this will be changed in gts 1.0, it's what `gts init` writes.

nodejs-server/package.json
alexander-fenster: Again, blaming `gts init`. Fixed!

nodejs-server/src/index.ts
alexander-fenster: Thanks! Fixed.

nodejs-server/src/echoServer.ts
alexander-fenster: I always forget about it! Fixed.

nodejs-server/src/echoServer.ts
alexander-fenster: (copying from another comment thread) I think (call, callback) => void is the interface that gRPC expects. This EchoServer class is being fed to grpc.Server.addService() and they expect certain API.

nodejs-server/src/operationsServer.ts
alexander-fenster: The new `gts` (tagged `1.0.0-0`) checks and fixes it.

alexander-fenster:
@JustinBeckwith @bcoe This is ready for another look!

Re: callbacks, I don't think I have a choice since gRPC server implementation wants me to provide functions with callbacks.

@noahdietz Are you OK with the folder location in this repository?

nodejs-server/src/echoServer.ts
```diff
@@ -0,0 +1,333 @@
...
+    }
+  }
+
+  /**
+   * Helper function to extract `google.rpc.Status` from some requests that
+   * contain this field.
+   */
+  private static requestToStatus(
+    request:
+      | showcaseV1Beta1.EchoRequest
```
bcoe: a bit of a meta question, about how we structure our TypeScript libraries in general. Will the default entry point of the library be whatever the current GA release of an API is. As an example, assume that we have API for a library that looks like this:

```
protos/google
   pubsub/v1
```

:point_up: when the library is first released, I assume our main `index.js` entry point would be the `pubsub/v1` version of the library ... now, 6 months passes, we release some amazing new features:

```
protos/google
   pubsub/v1
   pubsub/v2beta1
   pubsub/v2
```

Would the main entry point of the library become `pubsub/v2` once it's no longer in beta?

nodejs-server/src/echoServer.ts
bcoe: maybe not worth it, but could we wrap the `grpc.Server.addService` method, such that downstream libraries can use `async`, e.g.,

```js
addService(service) {
  grpc.Server.addService(util.callbackify(service))
}
```

maybe not worth it for this PR?

alexander-fenster:
@landrito This is one more implementation of Echo server, this time in TypeScript. I need this for several reasons to use in GAX Node.js CI (it would be much easier to use and control JS code, compared with using Docker or running binaries). Are you OK with the location of this code in the repo?

landrito:
@alexander-fenster This is awesome thanks! Just curious if there is no way to get the same ease of use from the current implementation? If not, lets make sure to add the context to the readme of `nodejs-server`. I am mostly on board with this code location but would like to think over it during lunch.

noahdietz:
I think this is great that we are able to implement the showcase server in other languages,  and we certainly learned from it (Alex found some bugs in gapic-generator-go and gcli while implementing this!).

I'm still not entirely sure that having multiple implementations helps the showcase cause in the long run. We'd have two implementations to update/maintain, one of them that implements only 1 of the several services. Multiple implementations could potentially fork the baseline confidence gained from showcase that all gapics conform to the agreed specification (i.e. there is a bug in 1 implementation used only by 1 generator, resulting in diff behavior than the other X generators using a common implementation). 

I just want to be really sure about the goals of duping the server implementation.

Personally, I'd like to clearly document the purpose of having another language implementation of the server.

alexander-fenster:
I understand the concerns of having several implementations to support. I can keep maintaining the Node version if we need it (add other services to support, etc.).

For now, this implementation solved two purposes for me: 

1. It proved to be easier (and surely more exciting) to write this code in TypeScript instead of downloading binaries and having a huge external binary dependency within TypeScript tests (it's technically possible, but it will cause complicated and system-dependent CI, and I would generally pay more for having in-language CI code compared with relying on binary dependencies).

2. Since we plan to bring more changes to GAX Node.js, it's very useful to be able to debug more pieces in-language (with having sources available and no need to recompile anything) and to have less black boxes. Again, that's not about being technically unable to do something, it's just about making things easier.

landrito:
What would be your strategy in the future, when showcase-proper has server-side validation in the future?

alexander-fenster:
As discussed with @landrito: merging this for now, will discuss more about our support of different implementations and support of the testing spec later this week on the summit.

PR #129, opened 2019-04-17

googleapis-publisher:


noahdietz:
As desired, the following changes are present:
* gapic:
  * fix for use of PageToken
  * addition of raw `Response` to iterator type
  * removal of non-idempotent retry config
* gcli:
  * respecting field_behavior annotation in flags
  * fix for consumption of page items in Paginated commands

PR #135, opened 2019-05-23
chore: update python wrapper image in circleci

noahdietz:


landrito:
Great would you mind creating a CL removing the kotlin smoke test so we can merge the annotations update?

PR #126, opened 2019-04-07
Update README.md

Inconnu08:
- grammar
- broken link

PR #139, opened 2019-05-23

googleapis-publisher:


noahdietz:
This looks like a dupe PR of the one that was merged, triggered accidentally by the merge of #126 .

I will look into why this happened, I thought it was supposed to only open these PRs on proto changes...

PR #147, opened 2019-07-08
Half-sentence in comment

LibanOdowa:
Not sure if you want to complete the sentence or get rid of it (I just deleted it for now)

### DIFF
--- a/schema/google/showcase/v1beta1/echo.proto
+++ b/schema/google/showcase/v1beta1/echo.proto
@@ -96,7 +96,8 @@ service Echo {
 }
 
 // The request message used for the Echo, Collect and Chat methods. If content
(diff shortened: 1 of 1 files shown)


schema/google/showcase/v1beta1/echo.proto
```diff
@@ -96,7 +96,7 @@ service Echo {
 }
 
 // The request message used for the Echo, Collect and Chat methods. If content
-// is set in this message then the request will succeed. If a status is
+// is set in this message then the request will succeed.
```
noahdietz: Can you please instead fix it to say: `If status is set in this message then the status will be returned as an error.`

LibanOdowa: Done.

PR #161, opened 2019-08-02
Enable Kotlin smoke test

### LINKED ISSUE #137: Reenable kotlin-smoke-test when it's updated
In #136 we disabled the `kotlin-smoke-test` job as a presubmit requirement. Once [gapic-generator-kotlin](https://github.com/googleapis/gapic-generator-kotlin) is updated with the new annotation configs and a new version is released, we should reenable the test here.

jbolinger:
Not sure what the original problem was, but hopefully this is all it takes. 

[resolves #137]

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,6 +11,8 @@ workflows:
           requires:
             - build
           filters: *all_commits
+      - kotlin-smoke-test:
+          filters: *all_commits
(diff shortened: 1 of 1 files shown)


noahdietz:
Thanks, this was disabled a while ago (when the annotations went through their final churn) and we just weren't sure how to fix the Kotlin stuff.

PR #174, opened 2019-08-08
Add Block method to Echo service

### LINKED ISSUE #66: Wait method uses LRO exclusively
The `wait` method in the `echo.proto` was changed to an LRO sometime after the 0.5 release.

The LRO method is useful, but removing the normal method prevents testing retry functionality. For context, waiting for an LRO vs. retrying a failed operation are distinct and non-related events in the Kotlin generator. This change allows me to test the former (+1) but no longer covers the later case. 

Could you consider adding the normal method in addition to the LRO one (i.e. a method that fails until the requested amount of time has elapsed?)

jbolinger:
This restores the original blocking `Wait` method. 

`Wait` was removed and replaced in #31 with a long running based `Wait` method. This implementation is copied from the original, but the method has been renamed to `Block` so that it doesn't interfere with the current `Wait` method.

[resolves #66]


noahdietz:
@alexander-fenster do we need to update the Node.js server impl in the same PR?

alexander-fenster:
@noahdietz I'm now using showcase binary in Node.js CI (thanks for @ramya-ramalingam's code that downloads and runs the binary) so I stopped using Node.js server. We can either update it now, or update it later, or never update it (and delete it since its only reason to exist was to use it in Node.js CI :) )

noahdietz:
Discussed offline: we will remove the Node.js server implementation in a separate PR.

PR #182, opened 2019-08-13
Add dummy LRO service

### LINKED ISSUE #180: Consider adding a dummy LRO service
Can we add a simple service that implements [operations.proto](https://github.com/googleapis/googleapis/blob/master/google/longrunning/operations.proto) as an optional part of showcase (not in echo). It doesn't need to actually manage operations but we'd like it to be available so we can exercise the generated helper methods for accessing these methods.

See https://github.com/googleapis/gapic-generator-ruby/issues/132 for more context.

jbolinger:
Very simple implementation of the remaining LRO methods.

Note that the LRO methods don't appear to be a part of the CLI and I did not attempt to add them.

[resolves #180]

### DIFF
--- a/server/services/operations_service.go
+++ b/server/services/operations_service.go
@@ -17,7 +17,10 @@ package services
 import (
 	"context"
 	"encoding/base64"
+	"math/rand"
+	"strconv"
 	"strings"
+	"time"
 
 	"github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/ptypes"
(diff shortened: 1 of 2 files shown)


server/services/operations_service.go
```diff
@@ -125,18 +128,70 @@ func searchFilterFunc(query string) func(b *pb.Blurb) bool {
 	}
 }
 
+// returns a successful response if the resource name is not blank
```
noahdietz: A godoc comment must begin with the method name

`// CancelOperation ...`

server/services/operations_service.go
```diff
@@ -125,18 +128,70 @@ func searchFilterFunc(query string) func(b *pb.Blurb) bool {
...
 func (s operationsServerImpl) CancelOperation(ctx context.Context, in *lropb.CancelOperationRequest) (*empty.Empty, error) {
-	return nil, status.Error(codes.Unimplemented, "google.longrunning.CancelOperation is unimplemented.")
+	if in.Name == "" {
+		return nil, status.Error(codes.NotFound, "cannot cancel operation without a name.")
+	}
+	return &empty.Empty{}, nil
 }
 
+// returns a fixed response matching the given PageSize if the resource name is not blank
```
noahdietz: Same thing regarding godoc comments needing to start with the method name.

server/services/operations_service.go
```diff
@@ -125,18 +128,70 @@ func searchFilterFunc(query string) func(b *pb.Blurb) bool {
...
+		})
+	}
+
+	return &lropb.ListOperationsResponse{
+		Operations:    operations,
+		NextPageToken: "txenNext",
+	}, nil
 }
 
+// returns a successful response if the resource name is not blank
```
noahdietz: Same thing regarding godoc comments needing to start with the method name.

server/services/operations_service.go
```diff
@@ -125,18 +128,70 @@ func searchFilterFunc(query string) func(b *pb.Blurb) bool {
...
 func (s operationsServerImpl) DeleteOperation(ctx context.Context, in *lropb.DeleteOperationRequest) (*empty.Empty, error) {
-	return nil, status.Error(codes.Unimplemented, "google.longrunning.DeleteOperation is unimplemented.")
+	if in.Name == "" {
+		return nil, status.Error(codes.NotFound, "cannot delete operation without a name.")
+	}
+	return &empty.Empty{}, nil
 }
 
+// randomly waits and returns an operation with the same name
```
noahdietz: Same thing regarding godoc comments needing to start with the method name.

server/services/operations_service.go
```diff
@@ -125,18 +128,70 @@ func searchFilterFunc(query string) func(b *pb.Blurb) bool {
...
 func (s operationsServerImpl) WaitOperation(ctx context.Context, in *lropb.WaitOperationRequest) (*lropb.Operation, error) {
-	return nil, status.Error(codes.Unimplemented, "google.longrunning.WaitOperation is unimplemented.")
+	if in.Name == "" {
+		return nil, status.Error(codes.NotFound, "cannot wait on a operation without a name.")
+	}
+
+	num := rand.Intn(500)
+	time.Sleep(time.Duration(num) * time.Millisecond)
```
noahdietz: Why not use the timeout in [lropb.WaitRequest](https://github.com/googleapis/go-genproto/blob/fa694d86fc64c7654a660f8908de4e879866748d/googleapis/longrunning/operations.pb.go#L400)? The page size was used in the List RPC. I have a feeling folks would expect the timeout to be respected. WDYT?

server/services/operations_service.go
jbolinger: TIL - thanks! done.

server/services/operations_service.go
jbolinger: done

server/services/operations_service.go
jbolinger: done

server/services/operations_service.go
jbolinger: done

server/services/operations_service.go
jbolinger: I don't really have a preference. The [`WaitOperation`](https://github.com/googleapis/googleapis/blob/2a2c5518f64010c4e458afc818e57ed24fecdf6d/google/longrunning/operations.proto#L116) is a little strange anyway. The docs basically say that it's free to ignore the timeout and it doesn't even have to wait. So, random seemed fitting to me.

noahdietz: Ok fine by me

PR #188, opened 2019-08-16
Remove next page token

jbolinger:
The token will cause clients to loop forever.


server/services/operations_service.go
```diff
@@ -164,7 +164,7 @@ func (s operationsServerImpl) ListOperations(ctx context.Context, in *lropb.List
 
 	return &lropb.ListOperationsResponse{
 		Operations:    operations,
-		NextPageToken: "txenNext",
+		NextPageToken: "",
```
noahdietz: Just remove this line entirely, it will default to an empty string.

jbolinger: Sure, done.

PR #199, opened 2019-09-04
release v0.5.0

noahdietz:


noahdietz:
> Just pointing out that having a version number in one place would've been better because it's easy to make a mistake when it needs to be updated in many places at once.

Yes you are right, but this repo automates it. Check RELEASING.md. There is a script that does it. `go run ./util/cmd/bump_version --minor`. 😄

PR #251, opened 2019-12-10
chore: improve comments on insecure channel use

### LINKED ISSUE #27: Update usage documentation.
Make needing to use an insecure channel to make requests to gapic-showcase more explicit.

noahdietz:
Fixes #27

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,12 +100,27 @@ $ gapic-showcase \
 must be used since the server does not implement auth. Client library generators with
 Showcase-based integration tests need to provide the insecure channel to the client library
 in the tests._**
+
 Example for Node.js:
 
 ```js
 const grpc = require('@grpc/grpc-js');
 const showcase = require('showcase');
 const client = new showcase.EchoClient({ grpc, sslCreds: grpc.credentials.createInsecure() });
+```
+
+Example for Go:
+
+```go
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -96,8 +96,10 @@ $ gapic-showcase \
   --user.email rumble@goodboi.com
 > name:"users/0" display_name:"Rumble" email:"rumble@goodboi.com" create_time:<seconds:1554414332 nanos:494679000 > update_time:<seconds:1554414332 nanos:494679000 >
 ```
-_* You can make requests to this server from your own client but an insecure channel
-must be used since the server does not implement auth._
+**_Note: You can make requests to this server from your own client but an insecure channel
+must be used since the server does not implement auth. Client library generators with
+Showcase-based integration tests need to provide the insecure channel to the client library
+in the tests._**
```
alexander-fenster: Maybe add some examples for a few languages? This suggestions costs like half an hour of browsing gRPC docs for those not directly familiar with the subject.

````suggestion
Example for Node.js:

```js
const grpc = require('@grpc/grpc-js');
const showcase = require('showcase');
const client = new showcase.EchoClient({ grpc, sslCreds: grpc.credentials.createInsecure() });
```
````

noahdietz: Great idea, and I'll add one for Go.

PR #335, opened 2020-04-15
fix: fix the typo in mtls logic

arithmetic1728:
I made a silly mistake, caught this problem when doing system tests.


noahdietz:
Sorry I should've caught that in review! Release to follow.

arithmetic1728:
> Sorry I should've caught that in review! Release to follow.

Thank you Noah! Sorry about that.

PR #357, opened 2020-04-30
WIP proto3 optional support

noahdietz:
This adds a new field to `Echo.EchoRequest` and `Echo.EchoResponse`:

    optional string presence = n;

(where `n` is the respective field number in each message)

This field utilizes the new `optional` keyword for proto3 field presence. The `echo_service` implementation of `Echo.Echo` copies the _exact_ value of `presence` on the incoming request message into the response message, echoing what it received.

This allows clients to test field presence by running tests with an unset and explicitly set request field.

Todo:

- [ ] update github.com/golang/protobuf version upon release
- [ ] update google.golang.org/protobuf version upon release
- [ ] update version of `protoc` used in CI upon release of protobuf
- [ ] update google.golang.org/protobuf version upon release
- [ ] temporarily disable micro-generator CI smoke tests until all "support" proto3_optional
- [ ] (maybe) add a method_signature with `presence` to enable testing of flattened methods


server/services/echo_service.go
```diff
@@ -46,7 +46,9 @@ func (s *echoServerImpl) Echo(ctx context.Context, in *pb.EchoRequest) (*pb.Echo
 		return nil, err
 	}
 	echoTrailers(ctx)
-	return &pb.EchoResponse{Content: in.GetContent()}, nil
+
+	// bypass the getter for EchoRequest.presence to preserve the exact field value
+	return &pb.EchoResponse{Content: in.GetContent(), Presence: in.Presence}, nil
```
alexander-fenster, line 51: just wondering - why `.GetContent()` but just `.Presence`? (you have a getter for `Presence` defined, should we use it?)

noahdietz: The comment I placed above this line explains.

The accessor `GetPresence` (see echo.pb.go diff) will return `""` if `Presence` is `nil`. So in order to preserve the _exact_ value, include whether it has been set or not, we must access the field directly, rather than use the `GetPresence` accessor

PR #374, opened 2020-05-12
feat: add non-slash resource patterns for `Blurb` message

xiaozhenliu-gg5:
1. add two non-slash resources in `Blurb` message, add two fields `legacy_room_id` and `legacy_user_id`.
2. update server side, for creating blurbs, if two new added fields are present, append the field to the resource name. 
```
// if  b.LegacyRoomId 
name := fmt.Sprintf("%s/blurbs/legacy/%s.%d", parent, b.LegacyRoomId, id)
// if  b.LegacyUserId 
name := fmt.Sprintf("%s/blurbs/legacy/%s~%d", parent, b.LegacyUserId, id)
```
3. In messsaging_service_test, create a blurb w/non-slash resource, get it from database and delete it.
4. add comment to the message and properly test the gapic-showcase for new changes.


server/services/messaging_service.go
```diff
@@ -303,7 +303,14 @@ func (s *messagingServerImpl) CreateBlurb(ctx context.Context, in *pb.CreateBlur
 	id := puid.Next()
 	name := fmt.Sprintf("%s/blurbs/%d", parent, id)
 	now := ptypes.TimestampNow()
-
+	if len(b.LegacyRoomId) >0 {
+		roomId := b.LegacyRoomId
+		name := fmt.Sprintf("%s/blurbs/%d.%s", parent, id, roomId)
```
noahdietz, line 308:
```suggestion
		name = fmt.Sprintf("%s/blurbs/%d.%s", parent, id, roomId)
```
In Go, `:=` declares a _new_ variable and assigns the value. We want to use just `=` so that we assign the new value to the already delcared `name` variable.

Declaring a variable and not using it is an compilation error in Go.

server/services/messaging_service.go
```diff
@@ -303,7 +303,14 @@ func (s *messagingServerImpl) CreateBlurb(ctx context.Context, in *pb.CreateBlur
...
+	}
+	if len(b.LegacyUserId) >0  {
+		userId := b.LegacyUserId
+		name := fmt.Sprintf("%s/blurbs/%d~%s", parent, id, userId)
```
noahdietz, line 312:
```suggestion
		name = fmt.Sprintf("%s/blurbs/%d~%s", parent, id, userId)
```

server/services/messaging_service.go
```diff
@@ -303,7 +303,14 @@ func (s *messagingServerImpl) CreateBlurb(ctx context.Context, in *pb.CreateBlur
 	id := puid.Next()
 	name := fmt.Sprintf("%s/blurbs/%d", parent, id)
 	now := ptypes.TimestampNow()
-
+	if len(b.LegacyRoomId) >0 {
+		roomId := b.LegacyRoomId
+		name := fmt.Sprintf("%s/blurbs/%d.%s", parent, id, roomId)
+	}
+	if len(b.LegacyUserId) >0  {
```
noahdietz, line 310:
```suggestion
	if b.GetLegacyUserId() != ""  {
```

server/services/messaging_service.go
```diff
@@ -303,7 +303,14 @@ func (s *messagingServerImpl) CreateBlurb(ctx context.Context, in *pb.CreateBlur
 	id := puid.Next()
 	name := fmt.Sprintf("%s/blurbs/%d", parent, id)
 	now := ptypes.TimestampNow()
-
+	if len(b.LegacyRoomId) >0 {
```
noahdietz, line 306:
```suggestion
	if b.GetLegacyRoomId() != "" {
```

server/services/messaging_service_test.go
```diff
@@ -361,6 +361,36 @@ func Test_Blurb_lifecycle(t *testing.T) {
...
+		&pb.GetBlurbRequest{Name: second.GetName()})
+	if err != nil {
+		t.Errorf("Get: unexpected err %+v", err)
+	}
+	if !proto.Equal(second, gotSecond) {
+		t.Error("Expected to get created blurb.")
+	}
+	_, err = s.DeleteBlurb(
+		context.Background(),
+		&pb.DeleteBlurbRequest{Name: second.Name})
```
noahdietz, line 390:
```suggestion
		&pb.DeleteBlurbRequest{Name: second.GetName()})
```

server/services/messaging_service.go
```diff
@@ -303,7 +303,14 @@ func (s *messagingServerImpl) CreateBlurb(ctx context.Context, in *pb.CreateBlur
...
+		roomId := b.LegacyRoomId
```
noahdietz, line 307:
```suggestion
		roomId := b.GetLegacyRoomId()
```

server/services/messaging_service.go
noahdietz, line 306: Go has this cool way of declaring/initializing a variable and then checking it in the scope of a conditional, we might want to use that here.

```go
if roomId := b.GetLegacyRoomId(); roomId != "" {
  name = fmt.Sprintf("%s/blurbs/%d.%s", parent, id, roomId)
}
```

Same goes for the `userId` conditional

server/services/messaging_service_test.go
xiaozhenliu-gg5: Thanks! I'm wondering what's the difference between `second.Name` and `second.GetName()` ? 
In the below delete test L408, it calls the Blurb's name as `&pb.DeleteBlurbRequest{Name: delete.Name})`🤓

xiaozhenliu-gg5:
Thanks for all the helpful feedback! @noahdietz Learned a lot of useful things about Go now :) Pretty interesting! 

I summarized all comments and made a commit to apply all suggesting changes (thank you!) and typescript-smoke-test fails b.c it does not support it now I think. I will make it work from that side and then update this PR.

server/services/messaging_service.go
xiaozhenliu-gg5: 👍

server/services/messaging_service.go
xiaozhenliu-gg5: Sounds cool! done.

xiaozhenliu-gg5: Got it, thanks!

server/services/messaging_service_test.go
noahdietz: As I mentioned earlier, we want to use the Getters to access the value of a message. This _can_ protect us from rogue `nil` pointer panics (e.g.  if `second` was `nil`, `GetName()` returns empty string rather than crashing as it would with a direct field access on a `nil`). The other thing is that since proto3_optional has been introduced, using the Getters protect us from accidentally breaking ourselves - i.e. adding `optional` to `Blurb.name` would change the underlying type to `*string` from `string`, which might not match the `DeleteBurbRequest.name` which isn't `optional` and is just a `string`. I hope that makes some sense.

The previously written code should use the Getter too, so let's not continue doing the more risky thing :) Don't worry about changing that code in this change, maybe we can refactor some stuff at a later date.

xiaozhenliu-gg5: Sounds very reasonable to me, changing it to Getters now, thanks!

schema/google/showcase/v1beta1/messaging.proto
```diff
@@ -317,6 +320,8 @@ message Blurb {
   google.protobuf.Timestamp update_time = 6 [
     (google.api.field_behavior) = OUTPUT_ONLY
   ];
+  string legacy_room_id = 7;
+  string legacy_user_id = 8;
```
noahdietz, line 324: Do you think these should be wrapped in a `oneof legacy_id`? That would avoid any confusion around what is supposed to happen if both are set. 

I'm not too worried about this case though, just an idea. FWIW it will change the implementation in the server/tests a bit (`oneof` fields are represented kind of weirdly in Go).

xiaozhenliu-gg5: Yes, I think it makes more sense to combine the two new added fields as oneof. It will be clearer for the users.
So for `name` and `user` fields in `Blurb`, one represents chat room resource name, and one as author (user), should we also make them as oneof? Thanks

noahdietz: > So for name and user fields in Blurb, one represents chat room resource name

Yeah the comments say something to that affect, but the server doesn't actually set the `user` field. I think that is there as an example of using a cross-proto resource_reference (User is defined in identity.proto). 

> should we also make them as oneof

Let's not change that.

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/googleapis/gapic-showcase
 require (
 	cloud.google.com/go v0.57.0
 	github.com/golang/protobuf v1.4.1
+	github.com/googleapis/gapic-generator-go v0.14.1 // indirect
```
noahdietz, line 6: Please revert this.

go.sum
```diff
@@ -120,6 +120,8 @@ github.com/google/pprof v0.0.0-20200212024743-f11f1df84d12/go.mod h1:ZgVRPoUq/hf
 github.com/google/pprof v0.0.0-20200229191704-1ebb73c60ed3/go.mod h1:ZgVRPoUq/hfqzAqh7sHMqb3I9Rq5C59dIz2SbBwJ4eM=
 github.com/google/pprof v0.0.0-20200430221834-fc25d7d30c6d/go.mod h1:ZgVRPoUq/hfqzAqh7sHMqb3I9Rq5C59dIz2SbBwJ4eM=
 github.com/google/renameio v0.1.0/go.mod h1:KWCgfxg9yswjAJkECMjeO8J8rahYeXnNhOm40UhjYkI=
+github.com/googleapis/gapic-generator-go v0.14.1 h1:nxw4C6nUS2JApiOeQWeUlkQr3RNKjwGpZBrV8ekjEWY=
```
noahdietz, line 123: Please revert this file.

go.mod
xiaozhenliu-gg5: done, thanks!

go.sum
xiaozhenliu-gg5: done

schema/google/showcase/v1beta1/messaging.proto
xiaozhenliu-gg5: sounds cool, changed the legacy_id to be `oneof` field. Thanks

server/services/messaging_service.go
```diff
@@ -303,7 +303,12 @@ func (s *messagingServerImpl) CreateBlurb(ctx context.Context, in *pb.CreateBlur
 	id := puid.Next()
 	name := fmt.Sprintf("%s/blurbs/%d", parent, id)
 	now := ptypes.TimestampNow()
-
+	switch legacyId := b.LegacyId.(type) {
+	case *pb.Blurb_LegacyRoomId:
+		name = fmt.Sprintf("%s/blurbs/%d.%s", parent, id, legacyId.LegacyRoomId)
+	case *pb.Blurb_LegacyUserId:
+		name = fmt.Sprintf("%s/blurbs/%d.%s", parent, id, legacyId.LegacyUserId)
```
noahdietz, line 310:
```suggestion
		name = fmt.Sprintf("%s/blurbs/%d~%s", parent, id, legacyId.LegacyUserId)
```

schema/google/showcase/v1beta1/messaging.proto
```diff
@@ -317,6 +320,18 @@ message Blurb {
...
+
+  // (-- aip.dev/not-precedent: This is designed for testing non-slash,
+  //     resource patterns --)
+  oneof legacy_id {
+    // The legacy id of the room. This field is used to trigger server
+    // to return a pattern of `rooms/{room_id}/blurbs/{blurb_id}.{legacy_room_id}`
+    string legacy_room_id = 7;
+
+    // The legacy id of the user. This field is used to trigger server
+    // to return a pattern of `rooms/{room_id}/blurbs/{blurb_id}~{legacy_user_id}`
```
noahdietz, line 332:
```suggestion
    // The legacy id of the user. This field is used to signal
    // the use of the compound resource pattern `rooms/{room_id}/blurbs/{blurb_id}~{legacy_user_id}`
```

schema/google/showcase/v1beta1/messaging.proto
```diff
@@ -317,6 +320,18 @@ message Blurb {
   google.protobuf.Timestamp update_time = 6 [
     (google.api.field_behavior) = OUTPUT_ONLY
   ];
+
+  // (-- aip.dev/not-precedent: This is designed for testing non-slash,
+  //     resource patterns --)
+  oneof legacy_id {
+    // The legacy id of the room. This field is used to trigger server
+    // to return a pattern of `rooms/{room_id}/blurbs/{blurb_id}.{legacy_room_id}`
```
noahdietz, line 328:
```suggestion
    // The legacy id of the room. This field is used to signal
    // the use of the compound resource pattern `rooms/{room_id}/blurbs/{blurb_id}.{legacy_room_id}`
```

schema/google/showcase/v1beta1/messaging.proto
```diff
@@ -317,6 +320,18 @@ message Blurb {
   google.protobuf.Timestamp update_time = 6 [
     (google.api.field_behavior) = OUTPUT_ONLY
   ];
+
+  // (-- aip.dev/not-precedent: This is designed for testing non-slash,
+  //     resource patterns --)
```
noahdietz, line 325:
```suggestion
  // (-- aip.dev/not-precedent: This is designed for testing non-slash
  //     resource patterns. Ordinarily, non-slash separators are discouraged. --)
```

schema/google/showcase/v1beta1/messaging.proto
```diff
@@ -289,7 +289,10 @@ message Blurb {
   option (google.api.resource) = {
     type: "showcase.googleapis.com/Blurb"
     pattern: "rooms/{room_id}/blurbs/{blurb_id}"
+    pattern: "rooms/{room_id}/blurbs/{blurb_id}.{legacy_room_id}"
     pattern: "user/{user_id}/profile/blurbs/{blurb_id}"
+    pattern: "user/{user_id}/profile/blurbs/{blurb_id}~{legacy_user_id}"
```
noahdietz, line 294: A section on Pattern Uniqueness is being added to the AIP and what we have proposed here violates the uniqueness principle: https://github.com/googleapis/aip/pull/498/files#diff-5dab22c31f3732f8150a18b7eeba7515

We should change the new, compound patterns to something else so that they are valid. @chrisdunelm any ideas?

chrisdunelm: Yes, this needs changing so every pattern is unique according to the uniqueness definition soon to be in AIP-123, as linked by @noahdietz.

I don't know the data model used by showcase, so not sure I can make any sensible suggestion about what to use.
Perhaps `rooms/{room_id}/blurbs/{blurb_id}/legacy/{legacy_room}.{legacy_blurb}` - although this may not make any sense!

Also, not to change now, but all these `_id` suffixes shouldn't really be here, as specified in AIP-123!

noahdietz: Thanks Chris, that suggestion helps. Maybe we should move the generated {blurb_id} out to the compound section, making the "legacy" a sub-collection of the parent's "blurbs" collection, like so:

```suggestion
    pattern: "rooms/{room_id}/blurbs/legacy/{legacy_room_id}.{blurb_id}"
    pattern: "user/{user_id}/profile/blurbs/{blurb_id}"
    pattern: "user/{user_id}/profile/blurbs/legacy/{legacy_user_id}~{blurb_id}"
```

> all these _id suffixes shouldn't really be here, as specified in AIP-123!

Great point. I made #376 to track this!

xiaozhenliu-gg5: Thanks for the advice, making the changes now.

schema/google/showcase/v1beta1/messaging.proto
```diff
@@ -289,7 +289,10 @@ message Blurb {
   option (google.api.resource) = {
     type: "showcase.googleapis.com/Blurb"
     pattern: "rooms/{room_id}/blurbs/{blurb_id}"
+    pattern: "rooms/{room_id}/blurbs/legacy/{legacy_room_id}.{blurb_id}"
```
chrisdunelm, line 292: This looks ok to me, although I'm no expert in showcase!

server/services/messaging_service.go
```diff
@@ -303,7 +303,12 @@ func (s *messagingServerImpl) CreateBlurb(ctx context.Context, in *pb.CreateBlur
 	id := puid.Next()
 	name := fmt.Sprintf("%s/blurbs/%d", parent, id)
 	now := ptypes.TimestampNow()
-
+	switch legacyId := b.LegacyId.(type) {
+	case *pb.Blurb_LegacyRoomId:
+		name = fmt.Sprintf("%s/blurbs/%d.%s", parent, id, legacyId.LegacyRoomId)
+	case *pb.Blurb_LegacyUserId:
+		name = fmt.Sprintf("%s/blurbs/%d~%s", parent, id, legacyId.LegacyUserId)
```
noahdietz, line 310: These format strings need to be updated to reflect the updated patterns.

xiaozhenliu-gg5: thank you! done.

schema/google/showcase/v1beta1/messaging.proto
xiaozhenliu-gg5: 👍

schema/google/showcase/v1beta1/messaging.proto
```diff
@@ -325,11 +325,11 @@ message Blurb {
   //     resource patterns. Ordinarily, non-slash separators are discouraged. --)
   oneof legacy_id {
     // The legacy id of the room. This field is used to signal
-    // the use of the compound resource pattern `rooms/{room_id}/blurbs/{blurb_id}.{legacy_room_id}`
+    // the use of the compound resource pattern `rooms/{room_id}/blurbs/{legacy_room_id}.{blurb_id}`
```
noahdietz, line 328:
```suggestion
    // the use of the compound resource pattern `rooms/{room_id}/blurbs/legacy/{legacy_room_id}.{blurb_id}`
```

schema/google/showcase/v1beta1/messaging.proto
```diff
@@ -325,11 +325,11 @@ message Blurb {
...
     string legacy_room_id = 7;
 
     // The legacy id of the user. This field is used to signal
-    // the use of the compound resource pattern `rooms/{room_id}/blurbs/{blurb_id}~{legacy_user_id}`
+    // the use of the compound resource pattern `rooms/{room_id}/blurbs/{legacy_user_id}~{blurb_id}`
```
noahdietz, line 332:
```suggestion
    // the use of the compound resource pattern `users/{user_id}/profile/blurbs/legacy/{legacy_user_id}~{blurb_id}`
```

PR #380, opened 2020-05-14
feat: add fields with proto3_optional

noahdietz:
Adds several primitive fields to `google.showcase.v1.User` that are marked with `proto3_optional`. Adds a `method_signature` with all of the `proto3_optional` fields.
Upgrades protoc version to v3.12.0-rc-2 in ci.

Chore: elide struct type name in slices in identity_service_test.go to remove linter warnings
Chore: disable `kotlin-smoke-test` job, it is not maintained and doesn't support `proto3_optional`
Chore: remove `go-smoke-test` job because it's redundant, gapic-generator-go is run in the `compile_protos` job
Chore: disable `python-smoke-test` and `typescript-smoke-test` jobs until they support `proto3_optional`, tracked by #382 and #383 respectively

Note: PR includes regenerated protobuf/gRPC API and CLI (no changes in Go GAPIC).

cc: @AlanGasperini @hkdevandla


server/services/identity_service_test.go
```diff
@@ -33,7 +33,7 @@ func Test_User_lifecycle(t *testing.T) {
 	first, err := s.CreateUser(
 		context.Background(),
 		&pb.CreateUserRequest{
-			User: &pb.User{DisplayName: "ekkodog", Email: "ekko@google.com"},
+			User: &pb.User{DisplayName: "ekkodog", Email: "ekko@google.com", Nickname: proto.String("ekko"), Age: proto.Int32(26)},
```
alexander-fenster, line 36: It might be a good time to start using `@example.com` instead of `@google.com` emails (see [here](https://www.iana.org/domains/reserved) for more information).

alexander-fenster:
re: TypeScript build failure - it requires https://github.com/googleapis/gapic-generator-typescript/pull/460 that I will merge within a day or two.

server/services/identity_service_test.go
noahdietz: I opened a cleanup task to refactor all uses of `google.com` to `example.com`: #381. There is another such cleanup task we will need to get to as well.

PR #426, opened 2020-07-23
feat: Add client-side retry/deadline testing surface

noahdietz:
The `SequenceService` enables client library generators to integration test the behavior of GAPIC client-side retry and deadlines.

Use of the surface in a gapic-showcase generator testing scenario is as follow:

1. The client defines a `Sequence` with a `Response` set for the server to respond with via `CreateSequence`
1. The client defines some retry and timeout settings for the generated `AttemptSequence` method
1. The client initiates the `AttemptSequence` method using the `Sequence.name` of the previously created `Sequence`
    a. The server will respond to each RPC attempt with the response status that correlates with the attempt, in the order they are received, adding response delay if defined in that attempt’s predefined `Response`
    b. The `Sequence` must have a terminal response or provoke termination via exceeding the deadline
1. Once the `AttemptSequence` invocation completes, the client retrieves the `SequenceReport` for the attempted `Sequence` and asserts that the proper number of `Attempts` were made and with the proper delay and deadline

Note: this PR contains the newly _generated_ code for the added `SequenceService` in the `client/` `cmd/gapic-showcase` and `server/genproto/` directories.

The following files are not generated and need reviewing:
* schema/google/showcase/v1beta1/sequence.proto
* server/services/sequence_service.go
* server/services/sequence_service_test.go


server/services/sequence_service_test.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	attempts := report.GetAttempts()
+	if len(attempts) != len(responses) {
+		t.Errorf("%s: expected number of attempts to be %d but was %d", t.Name(), len(responses), len(attempts))
+	}
+
+	d, _ := ctx.Deadline()
+	for n, a := range attempts {
+		if a.GetAttemptNumber() != int32(n) {
```
vchudnov-g, line 123: nit: here and below, I suggest `if got, want := ...; got !=  want {` to avoid code duplication

server/services/sequence_service_test.go
```diff
@@ -0,0 +1,223 @@
...
+	if c := status.Code(err); c != codes.NotFound {
+		t.Errorf("%s: expected error to be %s but was %s", t.Name(), codes.NotFound.String(), c.String())
+	}
+}
+
+func TestAttemptSequenceMissingName(t *testing.T) {
+	s := NewSequenceServer()
+	_, err := s.AttemptSequence(context.Background(), &pb.AttemptSequenceRequest{Name: ""})
+	if c := status.Code(err); c != codes.InvalidArgument {
+		t.Errorf("%s: expected error to be %s but was %s", t.Name(), codes.InvalidArgument.String(), c.String())
```
vchudnov-g, line 221: nit: Do you need to call `String()` explicitly in all these places? Don't the *f functions do that automatically if the arg satisfies the string interface?

server/services/sequence_service_test.go
noahdietz: Good idea! Done.

server/services/sequence_service.go
```diff
@@ -0,0 +1,178 @@
...
+	deadline, _ := ctx.Deadline()
+	dpb, err := ptypes.TimestampProto(deadline)
+	if err != nil {
+		return nil, status.Errorf(
+			codes.Internal,
+			err.Error(),
+		)
+	}
+
+	// Get the number of attempts, which correlates to this attempt's number.
```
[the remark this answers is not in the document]
noahdietz: That's clearer. Done.

server/services/sequence_service_test.go
noahdietz: Ah you are right, I removed them. Done.

PR #435, opened 2020-08-10

googleapis-publisher:


noahdietz:
Rejecting because it hasn't pulled the latest gapic-generator-go release

PR #508, opened 2020-12-02
Route REST requests to back-end services

vchudnov-g:
Each REST request is now routed to the appropriate back-end service based on its HTTP annotation path. The server now outputs a log message confirming the request was received. Note, however, that no request parameters are populated yet, so most requests FAIL (returning an error code back over REST responses). Populating request parameters comes next.

This adds a dependency on the gorilla/mux dispatcher.

**Running locally:**
To build and test:
`go install ./util/cmd/... && go run ./util/cmd/compile_protos && go test ./... && go install ./...`

To run the server, as usual:
`gapic-showcase run`

Local REST output:
```
$ curl http://localhost:7469/v1beta1/echo:echo  
{}

$ curl http://localhost:7469/v1beta1/echo:expand
ERROR: not implementing streaming methods yet                                                                                                                 

$ curl http://localhost:7469/v1beta1/users/someone/profile/blurbs
rpc error: code = NotFound desc = Parent  not found. 
```

**Review notes:**
- `server/genrest/` contains the generated files, which are large but may be useful for seeing what gets generated. The actual code is in other directories, mainly `util/genrest/`
- there are no tests for the generated files yet, since the content will change in upcoming PRs.
- FYI, I am trying a new workflow where I submit my changes to a branch of the master repo and create a PR from there. This will allow for having dependent PRs in GitHub (dependent PRs don't work across forks)


util/genrest/gomodelcreator.go
```diff
@@ -53,9 +53,11 @@ func NewGoModel(protoModel *protomodel.Model) (*gomodel.Model, error) {
...
 
 			restHandler := &gomodel.RESTHandler{
-				HTTPMethod:   binding.RESTPattern.HTTPMethod,
-				URIPattern:   binding.RESTPattern.Pattern,
-				PathTemplate: pathTemplate,
+				HTTPMethod:      binding.RESTPattern.HTTPMethod,
+				URIPattern:      binding.RESTPattern.Pattern,
+				PathTemplate:    pathTemplate,
+				StreamingServer: protoMethodDesc.GetClientStreaming(),
+				StreamingClient: protoMethodDesc.GetServerStreaming(),
```
software-dov, line 60: Are these inadvertently switched?

software-dov (commented):
First pass; much easier to read than the gorilla mux PR, but I definitely don't understand everything and will make a second pass to make sure I do a good job.

util/genrest/gomodelcreator.go
vchudnov-g: Yikes, yes! Good catch! Fixing that now.

software-dov (approved):
LGTM

server/genrest/testing.go
```diff
@@ -0,0 +1,242 @@
...
+import (
+	"context"
+	"net/http"
+
+	"github.com/golang/protobuf/jsonpb"
+	genprotopb "github.com/googleapis/gapic-showcase/server/genproto"
+)
+
+// HandleCreateSession translates REST requests/responses on the wire to internal proto messages for CreateSession
+//    Generated for HTTP binding pattern: /v1beta1/sessions
```
miraleung, line 29: General question: Will we ever have to handle non-slash resource name separators here?

util/genrest/goviewcreator.go
```diff
@@ -18,47 +18,130 @@ import (
...
+				continue
+			}
+
+			file.P("")
+			file.P("  var %s *%s.%s", handler.RequestVariable, handler.RequestTypePackage, handler.RequestType)
+			file.P("  // TODO: Populate %s with parameters from HTTP request", handler.RequestVariable)
+			file.P("")
+			// TODO: In the future, we may want to redirect all REST-endpoint requests to the gRPC endpoint so that the gRPC-registered observers get invoked.
+			file.P("  %s, err := backend.%sServer.%s(context.Background(), %s)", handler.ResponseVariable, service.ShortName, handler.GoMethod, handler.RequestVariable)
+			file.P("  if err != nil {")
```
miraleung, line 92: Just curious, what's the motivation for adding this literal block line-by-line rather than the entire subblock at once?

util/genrest/goviewcreator.go
```diff
@@ -18,47 +18,130 @@ import (
...
 		for _, handler := range service.Handlers {
 			handlerName := namer.Get("Handle" + handler.GoMethod)
-			file.P("//                For :%s", handler.URIPattern)
-			file.P("// Match URIs of form :%s", matchingPath(handler.PathTemplate))
-			file.P("func %s(w http.ResponseWriter, r *http.Request) {", handlerName)
-			file.P("  // serialize from r to: var %s %s.%s", handler.RequestVariable, handler.RequestTypePackage, handler.RequestType)
-			file.P("  %s := %s(%s)", handler.ResponseVariable, handler.GoMethod, handler.RequestVariable)
-			file.P("  // serialize back to w\n")
+			pathMatch := matchingPath(handler.PathTemplate)
+			registered = append(registered, &registeredHandler{pathMatch, handlerName})
```
miraleung, line 72: (Not for this PR)
Should we have integrated file-diffing tests so we can verify the expected and actual file outputs? Please ignore if these are already present in the codebase/PR somewhere.

miraleung (approved):

server/genrest/testing.go
vchudnov-g: I"m not sure. Do you have a reference about that? I haven't played much with resource names.

Regardless, URL paths have slash-separated components and variables are defined in terms of those, as per https://cloud.google.com/endpoints/docs/grpc-service-config/reference/rpc/google.api#path-template-syntax

util/genrest/goviewcreator.go
vchudnov-g: Primarily consistency, adhering to the convention used in gapic-generator-go and, I think, gapic-showcase.

You bring up a good point, though, that for multi-line literals we could use something like [dedent](https://github.com/lithammer/dedent)

Might be worth considering in the future; for now, I'm inclined to not add more changes.

util/genrest/goviewcreator.go
vchudnov-g: Maybe. TBH, the main value of having golden files IMO is to make it easier for reviewers to see how code changes affect the output; that's why I'm including the relevant generated files in this PR. Maybe what we should do instead of golden file tests *per se* is add a presubmit check that requires that if these `server/genrest/` auto-generated files would be changed by the code, that they get submitted in the same PR (so that they're reviewed)

PR #516, opened 2020-12-08
Capture URL path variables, and match HTTP verbs

vchudnov-g:
We now capture URL path variables. These are echoed in the server logs, but not yet populated in the request protos nor propagated to the backend services.

In addition, the URL matchers now get triggered when called with the correct HTTP verb (as configured in the HTTP annotation). Note that if a handled pattern is invoked with the wrong verb, no handler is called and no error is issued. We will have to figure out later how to detect and report/error on this.

In addition:
- We add checks to make sure that no recursive variables are defined in HTTP annotations, since this is actually disallowed by https://cloud.google.com/endpoints/docs/grpc-service-config/reference/rpc/google.api#path-template-syntax. I had missed the fine-print before :-(
- We remove the explicit Go formatting calls because that seems to be happening regardless


miraleung (approved):

server/genrest/echo.go
```diff
@@ -19,17 +19,28 @@ package genrest
...
+	urlPathParams := gmux.Vars(r)
+	numUrlPathParams := len(urlPathParams)
+
 	backend.StdLog.Printf("Received request matching '/v1beta1/echo:echo': %q", r.URL)
+	backend.StdLog.Printf("  urlPathParams (expect 0, have %d): %q", numUrlPathParams, urlPathParams)
+
+	if numUrlPathParams != 0 {
+		w.Write([]byte(fmt.Sprintf("unexpected number of URL variables: expected 0, have %d: %#v", numUrlPathParams, urlPathParams)))
+		return
+	}
```
software-dov, line 43: It looks like this block is repeated multiple times. How onerous would it be to lift it out into something named e.g. `verifyZeroPathParams`?

util/genrest/goviewcreator_test.go
```diff
@@ -40,8 +44,19 @@ func TestMatchingPath(t *testing.T) {
 			continue
 		}
 
-		if got, want := matchingPath(pathTemplate), testCase.expectMatch; got != want {
-			t.Errorf("testCase %2d: matchingPath error:\n    got: %q\n   want: %q", idx, got, want)
+		path, allVars, err := matchingPath(pathTemplate)
+
+		if got, want := (err != nil), testCase.expectError; got != want {
```
software-dov, line 49: I like this `got, want` pattern. It's very clear.

software-dov (approved):
LGTM, just one minor thing about the duplicated logic in all the handlers.

server/genrest/echo.go
vchudnov-g: This is a generated file, so I don't feel too bad about having repeated code. WDYT?

I suppose we could have the generated code call a common (non-generated) function to check for the expected number of params, but then the function call itself would be repeated. I'd like to wait until the generated functionality is more stable before doing that, because there's likely more similar refactoring we could do then.

I add a TODO for now so we don't forget about it.

Thoughts?

software-dov: Ah, if it's generated code then I agree, repeated chunks aren't too problematic.

PR #517, opened 2020-12-10
feat(genrest): populate scalar request fields with URL path variables

vchudnov-g:
We can now populate request fields with the values of URL path variables. The core functionality for this is extracted into a package `util/genrest/resttools` used by the generated code in `server/genrest` (and in the future, probably also by `util/genrest` itself for build-time checking).

With this core functionality in place, setting additional request parameters (body params and query params) will be pretty simple; upcoming PRs will do this.

Limitations: the data conversion works (and is tested) for scalar field types, even within nested messages, but not yet for lists or oneofs


util/genrest/resttools/populatefield.go
```diff
@@ -0,0 +1,140 @@
...
+		case protoreflect.StringKind:
+			protoValue = protoreflect.ValueOfString(value)
+		case protoreflect.BytesKind:
+			protoValue = protoreflect.ValueOfBytes([]byte(value))
+
+		case protoreflect.Int32Kind, protoreflect.Sint32Kind, protoreflect.Sfixed32Kind:
+			v64, err := strconv.ParseInt(value, 10, 32)
+			parseError, protoValue = err, protoreflect.ValueOfInt32(int32(v64))
+		case protoreflect.Uint32Kind, protoreflect.Fixed32Kind:
+			v64, err := strconv.ParseUint(value, 10, 32)
```
software-dov, line 104: Same here: `v32`?

util/genrest/resttools/populatefield.go
```diff
@@ -0,0 +1,140 @@
...
+		case protoreflect.FloatKind:
+			v64, err := strconv.ParseFloat(value, 32)
+			parseError, protoValue = err, protoreflect.ValueOfFloat32(float32(v64))
+		case protoreflect.DoubleKind:
+			v64, err := strconv.ParseFloat(value, 64)
+			parseError, protoValue = err, protoreflect.ValueOfFloat64(v64)
+
+		case protoreflect.BoolKind:
+			// TODO: should we be stricter in what we accept? ParseBool accepts various
+			// representations of "true" and "false" (https://golang.org/pkg/strconv/#ParseBool)
```
software-dov, line 123: I would argue it's less confusing and frustrating for the user if we were more strict initially and became less strict as needs arose. Then again, there isn't an obvious reason to be strict in this case.

software-dov (approved):
Looks good, just a minor naming convention question.

util/genrest/resttools/populatefield.go
```diff
@@ -0,0 +1,140 @@
...
+		// reference for proto scalar types:
+		// https://developers.google.com/protocol-buffers/docs/proto3#scalar
+
+		case protoreflect.StringKind:
+			protoValue = protoreflect.ValueOfString(value)
+		case protoreflect.BytesKind:
+			protoValue = protoreflect.ValueOfBytes([]byte(value))
+
+		case protoreflect.Int32Kind, protoreflect.Sint32Kind, protoreflect.Sfixed32Kind:
+			v64, err := strconv.ParseInt(value, 10, 32)
```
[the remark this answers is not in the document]
vchudnov-g: I meant for this to emphasize the static type, which here and below is a 64-bit numeric, even though if there's no error, the value is later cast to to a 32-bit type. If this is confusing I can just s/`v64`/`number`/g

util/genrest/resttools/populatefield.go
vchudnov-g: as above

util/genrest/resttools/populatefield.go
vchudnov-g: I'm leaving it as a TODO for now and then will investigate how strictly, if at all, Google and REST conventions enforce Boolean value representations in REST calls. That said, since Showcase is meant to be used primarily by GAPICs, we could be strict here to force GAPICs to be strict in what they emit. (If somebody else uses Showcase without GAPICs, they may have issues, but we don't necessarily need to optimize for them.)

util/genrest/resttools/populatefield.go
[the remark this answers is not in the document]
vchudnov-g: ok, I changed the naming for clarity.

PR #522, opened 2020-12-15
feat(genrest): Populate proto request fields from REST request body 

vchudnov-g:
We can now populate request fields with the values specified in REST request bodies, if they have been so configured by the HTTP annotations.

Limitations: This works fine when the HTTP annotation specifies `body:"*"` or `body:"foo"` where `foo` is itself a message; these seem to be the common cases. This has not yet been tested (and may not work) when `foo` refers to a scalar field; that will occur in an upcoming PR.

Review notes:
- `server/genrest/` contains generated files, so use them as reference but don't focus on those
- I suggest approaching this the logical call order:
  -  `util/genrest/protomodelcreator.go`
  -  `util/genrest/protomodel/**`
  - `util/genrest/gomodelcreator.go`
  -  `util/genrest/gomodel/**`
  - `util/genrest/goviewcreator*.go`


miraleung (approved):
Some general notes not specific to this PR:
- If all the code generation logic resides in `util/genrest`, would it make sense `genrest` to a top-level directory? Minimizing the classes of logic in `util` would help us avoid falling into the [util trap](https://darrenhobbs.com/2003/03/10/util-is-a-code-smell/).
- `genrest` appears to be generalizable to any proto, should we move this from showcase into a standalone tool? That way, we could test other APIs as well.

vchudnov-g:
Thanks, @miraleung! To answer your points:

* util trap: I don't think that quite applies here. The terminal package name is not `util` but `genrest`, and these really are utilities wrt the main purpose of the repo, which is to run a Showcase server or Showcase CLI. Various stuff under `util/` run as part of the release process or CI process. In particular, `util/cmd/compile_protos` and `util/genrest` create the auto-generated files that are part of the source code for Showcase, so thinking of them as tools serving the hand-written Showcase code makes sense IMO.
* generalizing `genrest`: I've thought about this before and I like the idea of generalizing it (and that _was_ a minor consideration in making it a `protoc` plugin), but I don't want us to get ahead of ourselves I mean this in two senses:
  1. we have a concrete use case for using it in Showcase, and we need it. Let's start with that and not postpone this by trying to make this general from the get-go. Generalization can come at a later iteration.
  2. I love the idea of having general tools, but let's make sure if we do release `genrest` separately, that there is a need/use case for it. Do people want/need a way to automatically generate a REST endpoint from their AIP-annotated proto3s? If so, let's definitely do it! But if not, let's resist premature generalization given that we have a whole bunch of goals we're trying to accomplish.
    - That said, I don't think it would be very hard to generalize, because this is a `protoc` plugin with very few Showcase-specific assumptions. There are some, though; for example, I would have to be more careful about guarding against namespace collisions, etc. So the work is not completely trivial. 
So let's get this working for Showcase first, and then consider whether we want to factor it out into its own module and how much effort that would take.

software-dov (approved):

PR #525, opened 2021-01-05
feat: add Bazel rules for referencing showcase protos

viacheslav-rostovtsev:


WORKSPACE.bazel
```diff
@@ -0,0 +1,68 @@
...
+  name = "com_google_googleapis",
+  strip_prefix = "googleapis-cf41866c6f14f10a07aa1e2a1260fc0a2727d889",
+  urls = ["https://github.com/googleapis/googleapis/archive/cf41866c6f14f10a07aa1e2a1260fc0a2727d889.zip"],
+  sha256 = "943602b1e91621f282c1e790eb46d9a70e202fe65091b7d41415a43973cebec4"
+)
+load("@com_google_googleapis//:repository_rules.bzl", "switched_rules_by_language")
+switched_rules_by_language(name = "com_google_googleapis_imports", gapic = True)
+
+##
+# protobuf for the protobuf protos
```
vchudnov-g, line 43: Can we simply update `gapic-showcase/schema/api-common-protos/WORKSPACE.bazel`?

WORKSPACE.bazel
```diff
@@ -0,0 +1,68 @@
...
+workspace(name = "gapic_showcase")
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+##
+# required by googleapis
+#
+http_archive(
+  name = "com_google_api_codegen",
+  strip_prefix = "gapic-generator-2.4.0",
+  urls = ["https://github.com/googleapis/gapic-generator/archive/v2.4.0.zip"],
```
vchudnov-g, line 27: Why do we need to depend on the monolith, if we're trying to deprecate it?

vchudnov-g (changes requested):
Let's make sure we're on the same page with @noahdietz and @vam-google  on how to structure this. 

For my own edification: This hasn't been a problem thus far with the µgenerators, right? What changes that we need to do this now?

WORKSPACE.bazel
viacheslav-rostovtsev: The showcase protos are not under the `api-common-protos` so we cannot use that `WORKSPACE.bazel` file for the showcase targets.

noahdietz:
Note: The codecov failure is just a flake, no need to worry about it.

@viacheslav-rostovtsev can you please share the use case, as Victor also asked? 

I see this being useful if we wanted to add GAPIC generation rules for Showcase somewhere, or as a test for a microgenerator project. Today, I think all microgenerator projects use either raw protos somehow or consume the fully resolved descriptor set included in the Release Assets. Not saying we shouldn't do this, just want to know what the motivation is.

WORKSPACE.bazel
viacheslav-rostovtsev: that was me using the wrong flag in the googleapis dependency
fixed

viacheslav-rostovtsev:
> This hasn't been a problem thus far with the µgenerators, right? What changes that we need to do this now?
Nobody really cared to have Bazel-addressable showcase protos before. The µgenerators are currently copying the showcase protos inside their repos (the Ruby and Typescript definitely do, others likely), and Bazel integration itself is fairly recent.
I was looking to generate a showcase library with the current changes and using Bazel seemed like a good way to achieve this so I added the bindings locally and now 'upstreaming' them.

Another usecase: Ruby-Bazel rules have an example workspace with usage demonstrations: https://github.com/googleapis/gapic-generator-ruby/tree/master/bazel_example/gapic-generator
With these bindings I would convert this example to use showcase instead of yet another `example.proto`.

WORKSPACE.bazel
vchudnov-g: Oh, right. 

(I suppose we could also put `.bazelrc` and `WORKSPACE` in `gapic-showcase/schema/`, but standard practice is to put it at the top of the repo so we can `BUILD` the repo. I'm fine with that. Thoughts, @noahdietz ?)

WORKSPACE.bazel
```diff
@@ -0,0 +1,59 @@
...
+  urls = ["https://github.com/googleapis/googleapis/archive/cf41866c6f14f10a07aa1e2a1260fc0a2727d889.zip"],
+  sha256 = "943602b1e91621f282c1e790eb46d9a70e202fe65091b7d41415a43973cebec4"
+)
+load("@com_google_googleapis//:repository_rules.bzl", "switched_rules_by_language")
+switched_rules_by_language(name = "com_google_googleapis_imports", grpc = True)
+
+##
+# protobuf for the protobuf protos
+#
+http_archive(
```
vchudnov-g, line 36: Can we get rid of `gapic-showcase/schema/api-common-protos/WORKSPACE` which defines essentially the same dependency (but older version) so we don't have nested workspaces?

viacheslav-rostovtsev: The `gapic-showcase/schema/api-common-protos` is another repository (https://github.com/googleapis/api-common-protos) brought by submodule so I don't think we **want** to change anything there.

Nested workspaces however are not an issue, the `bazel_example` in `gapic-generator-ruby` is a nested workspace (https://github.com/googleapis/gapic-generator-ruby/blob/master/bazel_example/WORKSPACE.bazel); it is a supported scenario.

vchudnov-g: Oh, I see. I misread that the `urls` line was referring to similar things. My bad.

I'm OK with this. Do please get sign-off from @noahdietz as well, though.

vchudnov-g (approved):
Do please get sign-off from @noahdietz as well, though.

noahdietz:
LGTM

PR #532, opened 2021-01-09
(feat) Provide gRPC transcoding test suite

vchudnov-g:
This provides a test suite (proto schema + proto-json data) that GAPIC generators can use to test their gRPC transcoding, This PR also contains tests to ensure the test suite is valid; that test can be used as a reference implementation by the generators.


schema/google/showcase/v1beta1/compliance.proto
```diff
@@ -0,0 +1,147 @@
...
+  repeated ComplianceGroup group = 1;
+}
+
+
+
+// DataPack is a message used for testing REST transcoding of
+// different data types. In the future, it may be part of an Echo
+// service RPC that also tests JSON responses.
+message DataPack {
+  DataPack subpack = 1;  // recursive message cause problems (infinite loop?) for go_cli
```
noahdietz, line 114: I can take a look at that. If we need to unblock this PR, can we just add this later?

schema/google/showcase/v1beta1/compliance.proto
```diff
@@ -0,0 +1,147 @@
...
+message ComplianceSuite {
+  repeated ComplianceGroup group = 1;
+}
+
+
+
+// DataPack is a message used for testing REST transcoding of
+// different data types. In the future, it may be part of an Echo
+// service RPC that also tests JSON responses.
+message DataPack {
```
noahdietz, line 113: Question: do we need this message if ComplianceData is the same thing sans recursive, nested reference?

util/execute.go
```diff
@@ -21,7 +21,10 @@ import (
 
 // Execute runs the given strings as a command.
 func Execute(args ...string) {
+	//	log.Printf(`running "%s"`, strings.Join(args, `" "`))
 	if output, err := exec.Command(args[0], args[1:]...).CombinedOutput(); err != nil {
 		log.Fatalf("%s", output)
+	} else {
+		// log.Printf("output:\n%s", output)
```
noahdietz, line 28: Nit: please remove these commented out bits.

cmd/gapic-showcase/endpoint_test.go
```diff
@@ -0,0 +1,110 @@
...
+			want = 200
+		}
+		if got := response.StatusCode; got != want {
+			t.Errorf("testcase %2d: status code: got %d, want %d", idx, got, want)
+			t.Errorf("  request: %v", request)
+		} else {
+			if want != 200 {
+				// we got the expected error
+				continue
+			}
```
noahdietz, line 96: Why not use an `else if`? I've reread this section and I can't figure out why it would be `if` nested in `else`, but I could totally be missing the desired case.

```suggestion
		} else if want != 200 {
			// we got the expected error
			continue
		}
```

noahdietz (commented):
Looking good to start!

schema/google/showcase/v1beta1/compliance.proto
vchudnov-g: I defined `DataPack` in an earlier PR as part of `echo.proto`, and I've been using it in other tests that are checked in. That's why I haven't removed it yet. 

I added `ComplianceData` in this PR to get around the recursion issue. If it can be fixed, then we certainly only need one of the two protos. 

Do you think the recursion issue is fixable? I haven't explored it.

I was using recursion as a simple way of testing that I can transcode nested fields. I can work around that by adding non-recursive sub-messages to to `ComplianceData`, and switch everything over to use that instead of `DataPack`

schema/google/showcase/v1beta1/compliance.proto
vchudnov-g: See above. I wanted to get this draft PR out as a POC, and existing tests are using `DataPack`

util/execute.go
vchudnov-g: Will do!

schema/google/showcase/v1beta1/compliance.proto
noahdietz: Oh ok, so there'd only need to be one of these "data" messages containing all of the field types?

schema/google/showcase/v1beta1/compliance.proto
noahdietz: Well, it could be fixed in the gen_cli, I'd just limit the depth of the recursion (probably to just one), but its a low priority thing and if we could have some simple separate message to test nested messages, then I think that would would be preferred. I think your suggestion of using non-recursive sub-message would be good, if you are OK with that.

vchudnov-g: Yeah, I'll do that.

schema/google/showcase/v1beta1/compliance.proto
vchudnov-g: Essentially, though I'd have to define something for the submessage.

cmd/gapic-showcase/endpoint_test.go
vchudnov-g: Done

util/execute.go
vchudnov-g: Done

schema/google/showcase/v1beta1/compliance.proto
vchudnov-g: Done in the commit I just pushed.

schema/google/showcase/v1beta1/compliance.proto
vchudnov-g: Done

noahdietz (commented):
LGTM! We'd want a compliance_service_test.go to cover everything (I know its a simple echo, but still). Do you want to take this out of Draft? Or open a diff PR?

vchudnov-g:
@noahdietz 
> We'd want a compliance_service_test.go to cover everything

`compliance_suite_test.go` should cover it, don't you think?

I will wait just a bit to take this out of draft mode, since I just pointed the team to the corresponding new design doc section--just in case there are objections.

Maybe in this PR or maybe in a follow-up, I will add a Compliance RPC that exercises path params.

noahdietz:
I was looking at the codecov comments on compliance_service.go and how sections of it aren't covered, which made me think there wasn't a sufficient test for it. You're right that the compliance_suite_test.go covers these handlers, so its probably not an issue.

vchudnov-g:
I rebased the PR onto master. It's ready for a formal review and subsequent merging.

server/services/compliance_service_test.go
```diff
@@ -0,0 +1,68 @@
...
+	for idx, rpc := range [](func(ctx context.Context, in *pb.RepeatRequest) (*pb.RepeatResponse, error)){
+		server.RepeatDataBody,
+		server.RepeatDataQuery,
+		server.RepeatDataSimplePath,
+	} {
+		response, err := rpc(context.Background(), request)
+		if err != nil {
+			t.Errorf("call %d: error: %s", idx, err)
+		}
+		if got, want := response.GetInfo(), request.GetInfo(); !proto.Equal(got, want) {
```
noahdietz, line 61: Nit: same comment here WRT use of `go-cmp`.

cmd/gapic-showcase/compliance_suite_test.go
```diff
@@ -0,0 +1,203 @@
...
+					continue
+				}
+				httpResponse, err := http.DefaultClient.Do(httpRequest)
+				if err != nil {
+					t.Errorf("%s error issuing call: %s", errorPrefix, err)
+					continue
+				}
+
+				// Check for successful response.
+				if got, want := httpResponse.StatusCode, 200; got != want {
```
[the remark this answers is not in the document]
vchudnov-g: Done

cmd/gapic-showcase/compliance_suite_test.go
```diff
@@ -0,0 +1,203 @@
...
+				// Unmarshal httpResponse body, interpreted as JSON.
+				// should do this.
+				responseBody, err := ioutil.ReadAll(httpResponse.Body)
+				httpResponse.Body.Close()
+				if err != nil {
+					t.Errorf("%s could not read httpResponse body: %s", errorPrefix, err)
+					continue
+				}
+				var response genproto.RepeatResponse
+				if err := jsonpb.UnmarshalString(string(responseBody), &response); err != nil {
```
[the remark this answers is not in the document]
vchudnov-g: Ooops, thanks for the catch! (I saw v1 was deprecated when yesterday, so I used `protojson` for what I added then, and meant to come back to the other files in a later PR. But I forgot to check for usages of v1 in this file!)

Done.

cmd/gapic-showcase/compliance_suite_test.go
```diff
@@ -0,0 +1,203 @@
...
+					continue
+				}
+
+				// Check for expected response.
+				if got, want := response.GetInfo(), requestProto.GetInfo(); !proto.Equal(got, want) {
+					gotString := jsonMarshaller.Format(got)
+					wantString := jsonMarshaller.Format(want)
+					t.Errorf("%s unexpected response:\n   -->got:\n`%s`\n   -->want:\n`%s`\n",
+						errorPrefix, gotString, wantString)
+				}
```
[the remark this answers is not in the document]
vchudnov-g: I can do that, but is the output of `cmp.Diff` with `proto.Equal` humanly readable. What I like this (in general, not so useful for this test suite) is that if we very `got` is correct, we can paste it as the expected value

cmd/gapic-showcase/compliance_suite_test.go
```diff
@@ -0,0 +1,203 @@
...
+	queryParams := []string{}
+	addParam := func(condition bool, key, value string) {
+		if !condition {
+			return
+		}
+		queryParams = append(queryParams, fmt.Sprintf("info.%s=%s", key, value))
+	}
+
+	if !exclude["f_string"] {
+		addParam(len(info.FString) > 0, "f_string", strings.ReplaceAll(strings.ReplaceAll(info.FString, " ", "+"), "%", "%%"))
```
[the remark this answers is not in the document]
vchudnov-g: Thanks for catch!

server/services/compliance_service_test.go
vchudnov-g: Done here, since the expected values are defined in Go. (Still assuming the diff output is readable)

cmd/gapic-showcase/compliance_suite_test.go
```diff
@@ -0,0 +1,201 @@
...
+	queryParams := []string{}
+	addParam := func(condition bool, key, value string) {
+		if !condition {
+			return
+		}
+		queryParams = append(queryParams, fmt.Sprintf("info.%s=%s", key, value))
+	}
+
+	if !exclude["f_string"] {
+		addParam(len(info.GetFString()) > 0, "f_string", strings.ReplaceAll(strings.ReplaceAll(info.FString, " ", "+"), "%", "%%"))
```
noahdietz, line 191: Nit: There is one more `info.FString` to replace in this line, at the end. Sorry! GitHub comments don't allow me to highlight sections of a line.

cmd/gapic-showcase/compliance_suite_test.go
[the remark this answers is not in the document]
noahdietz: Yeah the output is humanly readable. It produces (-/+) diffs of exactly what differed. This would be kind of bothersome to debug if broken because we'd have to "hand diff" the printed objects. I'd rather optimize for finding what is wrong. Up to you though, I'm not picky. :)

Updating "baseline" files or values is nice, but can be automated (we do this in gapic-generator-go via "update_golden" flag on the test suite) by having the test write the new value back to file.

cmd/gapic-showcase/compliance_suite_test.go
vchudnov-g: Ah, thanks!

cmd/gapic-showcase/compliance_suite_test.go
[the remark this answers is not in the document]
vchudnov-g: OK, let's try it.

schema/google/showcase/v1beta1/compliance_suite.json
```diff
@@ -0,0 +1,40 @@
...
+	  }
+	},
+	{
+	  "name": "Strings with spaces",
+	  "info": {
+            "f_string": "Hello there"
+	  }
+	},
+	{
+	  "name": "Strings with quotes",
```
miraleung, line 20: Do we care about weird edge cases like emojis?

vchudnov-g: We will. These are just very basic test cases for well-formed inputs. I plan to expand them in subsequent PRs.

noahdietz (approved):

PR #546, opened 2021-01-21
feat: Add more test cases to the gRPC-to-REST transcoding test suite

vchudnov-g:
This adds new RPCs to the Compliance service testing new ways of encoding request fields, via resource variable templates.

This also adds more test cases to test that conversions work as expected for a wider range of values and fields.

Known limitations:
- newlines `\n` in strings are not handled yet
- haven't verified the approach to encoding proto `bytes`


schema/google/showcase/v1beta1/compliance.proto
```diff
@@ -55,6 +55,18 @@ service Compliance {
       get: "/v1beta1/repeat/{info.f_string}/{info.f_int32}/{info.f_double}/{info.f_bool}:simplepath"
     };
   }
+
+  rpc RepeatDataPathResource(RepeatRequest) returns (RepeatResponse) {
+    option (google.api.http) = {
+      get: "/v1beta1/repeat/{info.f_string=first/*}/{info.f_child.f_string=second/*}/bool/{info.f_bool}:pathresource"
```
noahdietz, line 61: is the `bool` going to break things if https://github.com/googleapis/gapic-generator-go/pull/516 hasn't been merged or is it being replaced by the `sed` in compilation/generation script?

vchudnov-g: WDYM? `/bool/` is a literal part of the path. `{info.f_bool}` is a field name that will be replaced with the string value in that part of the URL and will then be cast to a bool.

When the Go generator is ready to use this test suite, it should use `%t` to emit `true` or `false`, and that should be parsed correctly here. That last part is covered in the `compliance_suite_test.go` and the `.json`

LMK if I'm missing your point.

noahdietz: Sorry, I meant that when `compile_protos` is run, and the Go GAPIC in the `client/`dir is regenerated, won't it see the `info.f_bool` field and generate code that does not compile? `url.QueryEscape(req.GetFBool())` where `GetFBool` returns a `bool` and `url.QueryEscape` expects a `string`. That is what the aforementioned PR fixes.

vchudnov-g: Oh, gotcha. No, I still have (ie, am not touching)  the workaround from the last PR that comments out that metadata insertion. All the tests pass locally.

(Sorry, I haven't included the auto-generated files for easier review; I like to add them last. If you want, I can add them earlier.)

noahdietz: As long as all of the generated code compiles, I am fine. If the workaround resolves the issue I mention too, then great!

vchudnov-g: Yes, it compiles. The workaround is https://github.com/googleapis/gapic-showcase/blob/master/util/compile_protos.go#L101, and we can remove it after the other PR gets merged.

vchudnov-g: I went ahead and added the auto-generated files to the PR.

noahdietz (approved):
LGTM but feel free to wait on other reviews

PR #550, opened 2021-01-22
chore: remove metadata replacement script

noahdietz:
With https://github.com/googleapis/gapic-generator-go/issues/514 fixed in generator version v0.17.2, we can remove this replacement script.


vchudnov-g (approved):
Oh, cool, thanks for doing this!

PR #553, opened 2021-01-23
chore: regen cmd/client/server directories

googleapis-publisher:


server/genrest/showcase-rest-sample-response.txt
```diff
@@ -118,7 +118,7 @@ Shim "Echo" (.google.showcase.v1beta1.Echo)
 ----------------------------------------
 Shim "Identity" (.google.showcase.v1beta1.Identity)
   Imports:
-    emptypb: "github.com/golang/protobuf/ptypes/empty" "github.com/golang/protobuf/ptypes/empty"
+    emptypbpb: "google.golang.org/protobuf/types/known/emptypb" "google.golang.org/protobuf/types/known/emptypb"
```
vchudnov-g, line 121: @noahdietz We have a `pbpb` stutter here, but it doesn't matter because this is mostly a file useful for dev and debugging. If `genrest` winds up using these imports we'll have to fix this, but I think we're fine for now.

At some point this file will go away. I won't block the approval on this.

vchudnov-g (approved):

PR #578, opened 2021-02-11
feat: handle+test single-message-field body REST transcoding; prevent duplication in query params; improve REST error handling

vchudnov-g:
- Adds the RPC `Compliance.RepeatDataBodyInfo` to test transcoding for RPCs annotated with a single `body` field
  - Configures the generator test suite to use this RPC
- Causes REST endpoints to generate errors if query parameters repeat body or path URI parameters. Real services may not error, but we want to ensure the generators are strict in what they emit.
  - Creates a Showcase-only test to verify this functionality
- Standardizes REST-level error handling to emit HTTP error codes and print uniform messages in the response and to the logs


vchudnov-g:
BTW, this PR does not yet include the autogenerated file changes it causes. I will add them before I merge, and can add them earlier if it's helpful for review. I just decided to not add them right away in order to not clutter the PR and file list.

cmd/gapic-showcase/compliance_suite_negative_test.go
```diff
@@ -0,0 +1,116 @@
...
+	suiteFile := filepath.Join(filepath.Dir(thisFile), "../../schema/google/showcase/v1beta1/compliance_suite.json")
+	jsonProto, err := ioutil.ReadFile(suiteFile)
+	if err != nil {
+		t.Fatalf("could not open suite file %q", suiteFile)
+	}
+	var suite pb.ComplianceSuite
+
+	if err := protojson.Unmarshal(jsonProto, &suite); err != nil {
+		t.Fatalf("error unmarshalling from json %s:\n   file: %s\n   input was: %s", err, suiteFile, jsonProto)
+	}
```
noahdietz, line 57: Can this stuff be moved into a function or two? Both this test file and the other use the same code (roughly the same I guess I haven't gone line-by-line). Maybe one to create a test server & return it (to then be started & the close deferred) and another for the compliance suite proto loading? Or one `setup` function that returns a server and the loaded suite?

cmd/gapic-showcase/compliance_suite_negative_test.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/googleapis/gapic-showcase/server/genproto"
+	pb "github.com/googleapis/gapic-showcase/server/genproto"
+	"google.golang.org/protobuf/encoding/protojson"
+)
+
+// TestComplianceSuiteErrors checks for non-spec-compliant HTTP requests. Not all of these
+// conditions necessarily generate a server error in a real service, but the behavior is often
+// ill-defined. We want Showcase to require the generators be strict in the transcoding format they
+// use.
+func TestComplianceSuiteErrors(t *testing.T) {
```
noahdietz, line 36: Nit: can we rename the file to `compliance_suite_errors_test.go` to match up with the main test function present? I think that is better name anyways, but I'm not too picky.

util/genrest/resttools/keysmatching.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// matchesSelfOrParent returns whether any element of `lookFor` is or contains a full or partial
+// path (in the dotted-field sense) to `examine`. In other words, this returns true when (a) the two
+// are identical, or (b) when `examine` starts with `lookFor` and is followed by a period.
+func matchesSelfOrParent(examine, lookFor string) bool {
+	if !strings.HasPrefix(examine, lookFor) {
+		return false
+	}
+	return len(examine) == len(lookFor) || examine[len(lookFor)] == '.'
```
noahdietz, line 46: An alternative:
```suggestion
	split := strings.Split(examine, ".")
	return split[0] == lookFor
```

This should cover the case where `lookFor` is the start of a dot-chain. If there is no dot, split returns the single string as is and you can do a direct comparison. This also does not return a false-positive when `lookFor` matches just a prefix of the entire field name in `examine`/`split[0]`.

vchudnov-g: I don't think this quite covers it. What you propose works when `examine==lookFor` and when `examine==lookFor+".foo.bar.baz"`, both as long as `lookFor` itself contains no `.`, but if  it does, these cases fail.  The second one should never happen because the proto body field should be a top-level field and the URL path params should be primitives, but the first case could happen with a dotted-path primitive in the path URL.

noahdietz: Indeed! You are right i did not think of `lookFor` having a `.`

noahdietz (approved):

PR #618, opened 2021-02-24
chore: draft v0.13.0 release

noahdietz:
Changes include:
- Add Compliance service for REGAPIC testing
- Add REGAPIC testing support with genrest
- Add mix-in service implementations
- Update API Service config with mix-ins and more
- Add Bazel proto_library targets for schema/
- Migrated to GitHub Actions
- Regen client & CLI with small updates
- Update dependencies

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Release History
 
+### v0.13.0 / 2021-02-24
+- Add Compliance service for REGAPIC testing
+- Add REGAPIC testing support with genrest
+- Add mix-in service implementations
+- Update API Service config with mix-ins and more
+- Add Bazel proto_library targets for schema/
+- Migrated to GitHub Actions
+- Regen client & CLI with small updates
+- Update dependencies
+
 ### v0.12.0 / 2020-08-12
 - Add client-side retry/deadline testing surface
 - Regen client & CLI with small updates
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # Release History
 
+### v0.13.0 / 2021-02-24
+- Add Compliance service for REGAPIC testing
+- Add REGAPIC testing support with genrest
```
vchudnov-g, line 5: I would reorder and rephrase these first two:
```
- Auto-generate REST endpoints for Showcase services (the genrest package)
- Add Compliance service for generators to use to test REST-transcoding their protos and RPCs
```

(REGAPIC is not a public name, so even though we're the main users of Showcase, it makes sense to minimize the use of that name in public repos and announcements)

I don't know whether it's worth adding `(not complete)` to these two bullets, since parts of the REST transcoding still have to be implemented (eg making PATCH calls work) and the test suite is not complete yet. On one hand it seems obvious; OTOH, it might be nice to communicate these features work but will be expanded. Maybe `(to be expanded in future releases)`?

vchudnov-g (approved):

CHANGELOG.md
noahdietz: Updated it with your wording and used `(partial)` on both bullets as discussed offline.

PR #626, opened 2021-03-01
chore: tidy go.sum

noahdietz:
For some reason, the [renovatebot PR](https://github.com/googleapis/gapic-showcase/pull/623/files#diff-3295df7234525439d778f1b282d146a4f1ff6b415248aaac074e8042d9f42d63) didn't properly tidy the `go.sum` which means some development tools don't work.

PR #627, opened 2021-03-01
chore: fix go vet/lint warnings

noahdietz:
The `lint` job fails [here](https://github.com/googleapis/gapic-showcase/pull/626/checks?check_run_id=2006650208) on `go vet` because of some usage of `testing` APIs from none `testing` goroutines. This fixes that and some other minor linter warnings.

Also, this tidies the `go.sum`, which was going to be done in https://github.com/googleapis/gapic-showcase/pull/626 but that was blocked by these vet warnings. The vet warnings fix failed CI b.c the `go.sum` needed to be fixed. So this submits them together.

### DIFF
--- a/server/services/messaging_service_test.go
+++ b/server/services/messaging_service_test.go
@@ -1518,6 +1518,8 @@ func TestConnect(t *testing.T) {
 				t.Fatal(err)
 			}
 		default:
+			// Empty default so that the select statement doesn't block
+			// and the loop can continue.
 		}
 	}
 
--- a/server/services/messaging_service_test.go
+++ b/server/services/messaging_service_test.go
@@ -268,8 +268,8 @@ func Test_UpdateRoom_invalid(t *testing.T) {
 
 func Test_UpdateRoom_alreadyPresent(t *testing.T) {
 	first := []*pb.Room{
-		&pb.Room{DisplayName: "Living Room"},
-		&pb.Room{DisplayName: "Weight Room"},
+		{DisplayName: "Living Room"},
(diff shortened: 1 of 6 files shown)


noahdietz:
Added @codyoss as a Go reviewer

server/services/messaging_service_test.go
```diff
@@ -1495,20 +1495,32 @@ func TestConnect(t *testing.T) {
 	}
 	s := NewMessagingServer(&mockIdentityServer{})
 
+	errs := make(chan error, 1)
```
codyoss, line 1498: This is fine, but there is also [errgroup](https://pkg.go.dev/golang.org/x/sync/errgroup)

server/services/messaging_service_test.go
```diff
@@ -1842,20 +1859,25 @@ func TestConnect_parentNotFoundLater(t *testing.T) {
 		context.Background(),
 		&pb.DeleteUserRequest{Name: first.GetName()})
 
+	err = <-errs
```
codyoss, line 1862: nit: could be one line

codyoss (approved):
LGTM, feel free to do with the comments as you will!

server/services/messaging_service_test.go
noahdietz: That's way better, I'll do that, thanks!

noahdietz: Done.

server/services/messaging_service_test.go
noahdietz: Done.

PR #630, opened 2021-03-01
chore: pin Go version in Ci

noahdietz:


vchudnov-g (commented):
Why pin the version as opposed to specifying a minimum version? Go make backwards-compatibility guarantees, right?

noahdietz:
> Go make backwards-compatibility guarantees, right?

Well, yes, except there was a change in the `go` tool that caused any command that builds the code to return an error if an adjustment to the go.mod is necessary. https://golang.org/doc/go1.16#go-command:

> Build commands like go build and go test no longer modify go.mod and go.sum by default. Instead, they report an error if a module requirement or checksum needs to be added or updated (as if the -mod=readonly flag were used). Module requirements and sums may be adjusted with go mod tidy or go get.

> Why pin the version as opposed to specifying a minimum version? 

Well it was setting a floor, but then it upgraded to this latest version and started reporting errors on some `go` commands. This is rather annoying and I can see this being an issue in the future with generated code changes.

vchudnov-g (approved):
Sigh. It's mildly annoying we have to pin it rather than floor it, but looks like we need it.

noahdietz:
> Sigh. It's mildly annoying we have to pin it rather than floor it, but looks like we need it.

I agree. We can change it in the future.

PR #652, opened 2021-04-01
DRAFT: real client example for interface abstraction primitives

software-dov:
Companion draft PR for https://github.com/googleapis/gapic-generator-go/pull/577
Ignore the metadata, that snuck in under the radar.


client/identity_client.go
```diff
@@ -90,7 +90,23 @@ func defaultIdentityCallOptions() *IdentityCallOptions {
...
+	CreateUser(context.Context, *genprotopb.CreateUserRequest, ...gax.CallOption) (*genprotopb.User, error)
+	GetUser(context.Context, *genprotopb.GetUserRequest, ...gax.CallOption) (*genprotopb.User, error)
+	UpdateUser(context.Context, *genprotopb.UpdateUserRequest, ...gax.CallOption) (*genprotopb.User, error)
+	DeleteUser(context.Context, *genprotopb.DeleteUserRequest, ...gax.CallOption) error
+	ListUsers(context.Context, *genprotopb.ListUsersRequest, ...gax.CallOption) *UserIterator
+}
+
+// IdentityClientHandle is a thin wrapper that holds an internalIdentityClient
+// It is agnostic as to the underlying transport, i.e. json+http, gRPC, or other.
+type IdentityClientHandle struct {
```
codyoss, line 104: What is the purpose of this handle? Should the interface(also?) be embeded in IdentityClient?

software-dov: This handle is intended to be a polymorphic wrapper around rest/grpc. The description in [](https://github.com/googleapis/gapic-generator-go/pull/577) describes the new style nomenclature in more detail, but the short description is:
* The old client is a newstyle, grpc backed 'transport'.
* The rest code will have a similar structure and also be considered a 'transport'.
* Exporting the common interface is frowned upon because adding new methods to it on API regeneration is technically a breaking change.
* The '*ClientHandle' struct is a 'new-style' client whose sole purpose is to wrap the interface. Because the interface is no longer exported, the technically breaking change is synchronized and no longer matters. Because *ClientHandle is a struct, adding new methods is a non-breaking change.

The factory functions for *ClientHandle are yet to be written. They will be named according to which kind of transport they construct and will take the corresponding parameters. The docs and samples will need to be updated to prefer new-style, *ClientHandle clients.

software-dov: To clarify: existing code will continue to work without modification. This is purely a set of changes made to facilitate abstraction.

codyoss: Sure but why can't IdentityClient itself just be polymorphic? Then expose a function to make it use a json+http backed transport?

software-dov: The only way I can imagine that working well is to change the client method implementations to have a runtime check to determine what the underlying transport mechanism is. That's gross to read and debug, complicates the implementation, makes maintenance harder. I agree that the higher order ClientHandle abstraction is a little wonky, but I think it gets the best of all worlds by
* not having additional runtime logic (barring dispatch, which is hidden)
* not breaking the existing surface
* not making the interface public, due to the aforementioned breaking change when adding methods

codyoss: I guess I am not quite following.

What I am picturing is 

```go
func IdentityClient {
   //...
  client internalIdentityClient
}

func NewIdentityClient(...) (*IdentityClient, error){
  //...
  return &IdentityClient{
     //...
    client: FactorcyGRPCClient(...)
  }
}

func NewJSONIdentityClient(...) (*IdentityClient, error){
  //...
  return &IdentityClient{
     //...
    client: FactorcyJSONClient(...)
  }
}

func (c *IdentityClient) CreateUser(...) (*genprotopb.User, error) {
  return c.client.CreateUser(...)
}
```

Am I missing something?

software-dov: No, that's correct; that is what the intended surface is to look like.
I'm hesitant to repurpose the existing Client structure for a few reasons. The primary one is the `CallOptions` field; it can't be removed because it's exported, and so it would either need to be duplicated in the subordinate structs or be explicitly passed into methods. This complicates the interaction between the 'wrapper' struct and the 'implementation' struct and requires explicit boilerplate that was previously implied. It also complicates the development process somewhat to be moving and renaming instead of just adding; this could be handled via development as is, as a separate branch to avoid contaminating the generated surface with unremovable structure definitions, followed by a refactor.
If you feel strongly about not muddying the waters for the users and for repurposing the existing Client structure, I'll change the gutting/addition of structures, but I'm definitely still in the camp of making the ClientHandle a new struct.

software-dov: Yes, I think we would need call options for REST as well, and I think you're right: specific call options are indeed not transport agnostic. My point is that the field, and the type, and the corresponding function signature changes, make the movement tricky and verbose.

tbpg: I'm afraid I don't fully understand what `IdentityClientHandle` (and the other `Handle` types) are for.

Cody's sample code above looks good to me, and what I would expect here. I don't know how to fit that in with the `Handle` types.

software-dov: There is only the one `Handle` type. It is intended to be a thin wrapper to
1) Abstract away the differences between grpc and rest transports
2) Prevent breaking changes when adding new methods to an API.

Its design was inspired by the C++ PIMPL idiom; the sole functionality is to wrap an interface.

tbpg: > C++ PIMPL idiom

TIL. Very interesting.

It might help me if I saw how `IdentityClientHandle` was used directly. Do users need to know about `IdentifyClientHandle`? How would they create one? Right now, I'm leaning toward reusing the existing Client type. But, I am not sure I understand all of the necessary features that are coming.

software-dov: Of course, an example is always helpful. Some details are handwaved or strawmanned, especially the factory names.

```go
var client IdentifyClientHandle

if transport == "grpc" {
    client = NewIdentifyGrpcClient(ctx)
} else if transport == "rest" {
    client = NewIdentifyRestClient(ctx)
} else if transport == "make_factory_decide" {
    client = NewIdentifyClientFromArgs(ctx, args)
} else {
    return fmt.Error("unknown transport: %s", transport)
}

 // Same name, same sig as current
response, err := client.Identify(ctx, request) 
```

The following would still be valid, just deprecated:
```go
var client IdentifyClient
client = NewIdentifyClient(ctx)
response, err := client.Identify(ctx, request)
```

codyoss: I am in favor of not introducing this new `Handle` type. I don't think it provides much value to the end-user if we could instead just repurpose "IdentifyClient" to do the same thing today. This would also allow our current users to not need to do anything or get worried about a deprecation.

software-dov: Generated surface has been changed and revised in light of the above discussion.

noahdietz:
Note: if you want to reduce some diffs, you can regenerate this based on master and the mixins RPCs will be dropped from the diff - they were merged today.

software-dov:
Updated to remove extraneous diffs.

client/compliance_client.go
```diff
@@ -175,7 +169,7 @@ type complianceGrpcClient struct {
 // This service is used to test that GAPICs can transcode proto3 requests to
 // REST format correctly for various types of HTTP annotations.
 func NewComplianceClient(ctx context.Context, opts ...option.ClientOption) (*ComplianceClient, error) {
-	clientOpts := defaultComplianceGrpcCallOptions()
+	clientOpts := defaultComplianceGrpcClientOptions()
```
codyoss, line 172: nit: `Grpc` should be `GRPC`. Other examples: [transport.DialGRPC](https://pkg.go.dev/google.golang.org/api/transport#DialGRPC) [option.WithGRPCDialOption](https://pkg.go.dev/google.golang.org/api/option#WithGRPCDialOption)

client/echo_client.go
```diff
@@ -213,33 +240,33 @@ func NewEchoClient(ctx context.Context, opts ...option.ClientOption) (*EchoClien
 
 	c.locationsClient = locationpb.NewLocationsClient(connPool)
 
-	return c, nil
+	return &EchoClient{internalEchoClient: c, CallOptions: callOpts}, nil
```
vchudnov-g, line 243: Let's be careful here: `&(EchoClient.CallOptions) != &callOptions`, right? After this line, we could say `EchoClient.CallOptions = altCallOptions` and then have `*(c.callOpts) != altCallOptions`, right?

I think you want 
```
echoClient = &EchoClient{internalEchoClient: c, CallOptions: callOpts}
c.CallOptions = &EchoClient.CallOptions
return echoEclient
```
Feel free to check me.
```

File: client/echo_client.go [hunk 4]
```diff
@@ -261,9 +291,8 @@ func (c *EchoClient) Echo(ctx context.Context, req *genprotopb.EchoRequest, opts
 
 // Expand this method splits the given content into words and will pass each word back
 // through the stream. This method showcases server-side streaming RPCs.
-func (c *EchoClient) Expand(ctx context.Context, req *genprotopb.ExpandRequest, opts ...gax.CallOption) (genprotopb.Echo_ExpandClient, error) {
+func (c *echoGrpcClient) Expand(ctx context.Context, req *genprotopb.ExpandRequest, opts ...gax.CallOption) (genprotopb.Echo_ExpandClient, error) {
 	ctx = insertMetadata(ctx, c.xGoogMetadata)
-	opts = append(c.CallOptions.Expand[0:len(c.CallOptions.Expand):len(c.CallOptions.Expand)], opts...)
```
noahdietz, line 266: This needs to be replaced.

client/echo_client.go
```diff
@@ -188,15 +215,18 @@ func NewEchoClient(ctx context.Context, opts ...option.ClientOption) (*EchoClien
...
-
-		echoClient: genprotopb.NewEchoClient(connPool),
+		echoClient:       genprotopb.NewEchoClient(connPool),
+		CallOptions:      &client.CallOptions,
 	}
 	c.setGoogleClientInfo()
 
+	client.internalEchoClient = c
+
 	c.LROClient, err = lroauto.NewOperationsClient(ctx, gtransport.WithConnPool(connPool))
```
noahdietz, line 230: I think the same thing that happened with the CallOptions needs to happen to the `LROClient` - the wrapper client should get the initially created one, and the internal client should get a reference to it.

client/echo_client.go
noahdietz: Also, we will need to update the generator to handle the extra "deref" of the `**CallOptions`.

client/compliance_client.go
```diff
@@ -141,51 +162,54 @@ func NewComplianceClient(ctx context.Context, opts ...option.ClientOption) (*Com
...
+	client.internalComplianceClient = c
+
 	c.operationsClient = longrunningpb.NewOperationsClient(connPool)
 
 	c.iamPolicyClient = iampb.NewIAMPolicyClient(connPool)
 
 	c.locationsClient = locationpb.NewLocationsClient(connPool)
 
-	return c, nil
+	return &client, nil
```
vchudnov-g, line 183: nit: for the three assignments above: `c.fooClient = ...`: does it make sense to put them inside the literal `c := &complianceGRPCClient{...}` definition for brevity/clarity?

software-dov: Setting them separately was the convention in the existing code, but I agree, it's cleaner.

software-dov:
The corresponding generator PR has been merged; this PR will be closed.

PR #662, opened 2021-04-13
chore: fix collision with operation helper

noahdietz:
There is an RPC named `Echo.Wait` that returns an LRO which creates a helper `WaitOperation` which is the same name as the RPC `Operations.WaitOperation` generated as a mixin method. This is a pretty one-off case as a method named `Wait` returning an LRO is not likely in an actual API. So we just rename the generated mixin method using `sed`  post generation.

### DIFF
--- a/util/compile_protos.go
+++ b/util/compile_protos.go
@@ -96,6 +96,16 @@ func CompileProtos(version string) {
 			"cmd/gapic-showcase/wait.go",
 			"s/EndEnd_time/EndEndTime/g",
 		},
+		// Fixes a name collision with the operation helper WaitOperation by renaming the mixin method.
+		{
+			"client/echo_client.go",
+			"1,/WaitOperation(ctx/{s/WaitOperation(ctx/WaitOperationMixin(ctx/;}",
+		},
+		// Fixes a name collision with the operation helper WaitOperation by renaming the mixin method.
+		{
+			"client/echo_client_example_test.go",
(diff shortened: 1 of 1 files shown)


viacheslav-rostovtsev (approved):

vchudnov-g:
If this is meant to be a "showcase" of how GAPICs work, having this post-processing seems less than ideal. Maybe we should create an issue to change the `Echo` rpc to something like `Delay` or `TimeOut` (and we'd have to change any clients using it, which I admit is hard), and then revert this PR. That way people can see how to use mixins properly and how they work/are named. Thoughts?

vchudnov-g:
Additionally, we should come up with a name to deal with (the admittedly rare) naming conflicts, which might simply be erroring during generation.

noahdietz:
The `Echo` service is decidedly non-standard, but I agree we could rename this one RPC to make it conform better. The `Echo` service, imo, has always been a testing surface and not an example one. Those are `Identity` and `Messaging`. WDYT?

vchudnov-g:
I think renaming the RPC makes sense. If it's not the main showcasing API, it makes it less onerous to fix (presumably, only breaking ourselves), but also less urgent. In the spirit of clarity/hygiene, I say we add a task to fix this, which would also mean making sure any generators using these RPCs get fixed.

noahdietz:
> In the spirit of clarity/hygiene, I say we add a task to fix this, which would also mean making sure any generators using these RPCs get fixed.

Indeed, I will do so!

PR #696, opened 2021-05-03
feat: Enforce checking for optional field presence/absence in test suite requests

vchudnov-g:
**Note** that this uses the `embed` package, which requires a switch to Go 1.16.

The `compliance_suite.json` file is moved from its old location under `schema/` to a location under `server` because `embed` does not follow symlinks. The original location under `schema/` becomes a symlink to the new location so that we may continue easily releasing the suite with other `schema/` files.


noahdietz:
> **Note** that this uses the `embed` package, which requires a switch to Go 1.16.

You will need to update the version of Go used in all GitHub Actions.

vchudnov-g:
Update the Go version in a bunch places. Have I missed any?

noahdietz:
> Update the Go version in a bunch places. Have I missed any?

lgtm! we will find out if we missed some ;)

noahdietz:
Looks like there was a linter violation. You can check the linter results from the Actions UI, and you'll find the "Artifacts" dropdown in the top right has a `linting-results` file.

vchudnov-g:
Oops, missed that! Should be fixed now.

vchudnov-g:
I'll add some tests to bring the coverage back up, but please do start reviewing. ;-)

noahdietz (approved):
lgtm once you get the added coverage :)

vchudnov-g:
Yay! I got the code coverage up to 100% on this patch!

PR #706, opened 2021-05-06
chore: fix multiline truncation of release notes

### LINKED ISSUE #691: GitHub action for release only captures first line of the release notes in CHANGELOG.md
I noticed this when pushing the v0.14.0 release, which only grabbed the first line of the notes for that release from `CHANGELOG.md`. I've since manually corrected the release notes for that one release, but we should figure out how to fix the process.

noahdietz:
In order to capture multiline strings in `set-output` we need to escape the new line characters prior to providing it to `set-output`. I tested the commands to assign & parse the notes, but we will just have to see what happens with the Action on the next release.

Fixes #691

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -24,7 +24,10 @@ jobs:
         run: echo ::set-output name=raw_version::"${GITHUB_REF#refs/*/v}"
       - name: Get release information
         id: changes
-        run: echo ::set-output name=release_notes::"$(go run ./util/cmd/print_changes ${{ steps.tag.outputs.version }})"
+        # Need to escape new line characters in order to get multiline release notes.
+        run: |
+          notes=$(go run ./util/cmd/print_changes ${{ steps.tag.outputs.version }})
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -24,7 +24,10 @@ jobs:
         run: echo ::set-output name=raw_version::"${GITHUB_REF#refs/*/v}"
       - name: Get release information
         id: changes
-        run: echo ::set-output name=release_notes::"$(go run ./util/cmd/print_changes ${{ steps.tag.outputs.version }})"
+        # Need to escape new line characters in order to get multiline release notes.
+        run: |
+          notes=$(go run ./util/cmd/print_changes ${{ steps.tag.outputs.version }})
+          echo ::set-output name=release_notes::"${notes//$'\n'/'%0A'}"
```
vchudnov-g, line 30: nice shell foo!
For my reference:
https://tldp.org/LDP/abs/html/string-manipulation.html
http://www.gnu.org/software/bash/manual/html_node/ANSI_002dC-Quoting.html

vchudnov-g (approved):

PR #713, opened 2021-05-13
chore: disable codecov

noahdietz:
I thought I disabled this earlier, but I guess I didn't. I will roll any secrets in this repo again to be safe.


vchudnov-g (approved):
Why were we doing this? It was slowing down our development?

noahdietz:
> Why were we doing this? It was slowing down our development?

Codecov had a security incident that leaked any information in the Action environment, potentially including Secrets provided via the environment. Ping me offline for the details if you are interested.

PR #714, opened 2021-05-14
feat: Require incoming REST requests to have correct x-goog-api-client header

vchudnov-g:
Showcase now requires incoming REST requests to have, in the header "x-goog-api-client", a token beginning with "rest/" and a token beginning with "gapic/"

This includes tests of this enforcement, and tighter (error-specific) tests for the content-encoding header enforcement as well.

### DIFF
--- a/util/genrest/resttools/checkrequestformat_test.go
+++ b/util/genrest/resttools/checkrequestformat_test.go
@@ -46,7 +46,7 @@ func TestCheckRequestFormat(t *testing.T) {
 		label     string
 		json      string
 		header    http.Header // nil means standard headers
-		wantError bool
+		wantError string
 	}{
 		{
 			label: "normal case",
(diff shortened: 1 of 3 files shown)


util/genrest/resttools/headers.go
```diff
@@ -15,19 +15,70 @@
...
+	var haveREST, haveGAPIC bool
+	for _, token := range strings.Split(content[0], " ") {
+		trimmed := strings.TrimSpace(token)
+		if !haveREST && strings.HasPrefix(trimmed, headerValueTransportRESTPrefix) {
+			haveREST = true
+		} else if !haveGAPIC && strings.HasPrefix(trimmed, headerValueClientGAPICPrefix) {
+			haveGAPIC = true
+		} else {
+			// nothing changed
+			continue
```
noahdietz, line 71: Is this else case necessary? I haven't thought through all of the scenarios, but thought I'd ask.

util/genrest/resttools/checkrequestformat_test.go
```diff
@@ -108,8 +157,16 @@ func TestCheckRequestFormat(t *testing.T) {
 		if request.Header == nil {
 			PopulateRequestHeaders(request)
 		}
-		if err := CheckRequestFormat(request.Body, request.Header, complianceData.ProtoReflect()); (err != nil) != testCase.wantError {
-			t.Errorf("test case %d[%q] text enum encoding: expected error==%v, got: %v", idx, testCase.label, testCase.wantError, err)
+		err = CheckRequestFormat(request.Body, request.Header, complianceData.ProtoReflect())
+		if (err != nil) != (len(testCase.wantError) > 0) {
```
noahdietz, line 161: `(testCase.wantError != "")`?

noahdietz (approved):
LGTM

util/genrest/resttools/headers.go
vchudnov-g: Not necessary, but a pedantic optimization. We could have just three linear if-statements and things would work fine, if "inefficiently". But then I decided to be more explicit (which might be clearer or muddier, not sure): if `haveREST` is not set yet, see whether we can set it; otherwise if `haveGAPIC` is not set yet, see whether we can set it; otherwise, we didn't change either, so continue with the loop (this else case). Fall-through: one of the two variables was just set, so let's see whether we can exit the loop prematurely.

(The efficiency point is just a principle thing, not practical; we're only dealing with a handful of tokens in the header.)

util/genrest/resttools/checkrequestformat_test.go
vchudnov-g: Uhm, yeah, that. Thanks for the catch!

PR #751, opened 2021-06-14
draft(WIP): regeneration of showcase client with query param changes

software-dov:
Companion PR, **NOT** intended to be merged.

This PR describes the surface changes as made by https://github.com/googleapis/gapic-generator-go/pull/659


client/compliance_client.go
```diff
@@ -585,6 +654,849 @@ func (c *complianceGRPCClient) CancelOperation(ctx context.Context, req *longrun
...
+func (c *complianceRESTClient) RepeatDataBodyInfo(ctx context.Context, req *genprotopb.RepeatRequest, opts ...gax.CallOption) (*genprotopb.RepeatResponse, error) {
+	m := protojson.MarshalOptions{AllowPartial: true, EmitUnpopulated: true}
+	jsonReq, err := m.Marshal(req)
+	if err != nil {
+		return nil, err
+	}
+
+	body := req.GetInfo()
+	queryParamStrs := []string{}
+	if body.GetFBool() {
```
vchudnov-g, line 729: So `RepeatDataBody` has, in it's `google.api.http` annotation, `body: "*"`. Thus, it should have no query params. See https://cloud.google.com/endpoints/docs/grpc-service-config/reference/rpc/google.api#rules-for-http-mapping

vchudnov-g: minor correction: My comment is actually for `RepeatDataBody` (which is the one that has `body: "*"`), i.e. the previous method, which should have no query params.

Here, since all the query params you compute are included inside the `RepeatDataBodyInfo`'s `body: "info"` field, they shouldn't be query params either.

Refer also to https://google.aip.dev/127

client/compliance_client.go
```diff
@@ -585,6 +654,849 @@ func (c *complianceGRPCClient) CancelOperation(ctx context.Context, req *longrun
...
+func (c *complianceRESTClient) RepeatDataQuery(ctx context.Context, req *genprotopb.RepeatRequest, opts ...gax.CallOption) (*genprotopb.RepeatResponse, error) {
+	m := protojson.MarshalOptions{AllowPartial: true, EmitUnpopulated: true}
+	jsonReq, err := m.Marshal(req)
+	if err != nil {
+		return nil, err
+	}
+
+	queryParamStrs := []string{}
+	if req.GetInfo() != nil {
+		queryParamStrs = append(queryParamStrs, fmt.Sprintf("info=%v", req.GetInfo()))
```
vchudnov-g, line 850: Query params should be leaf fields. 

https://cloud.google.com/endpoints/docs/grpc-service-config/reference/rpc/google.api#rules-for-http-mapping:
> parameter name is the field path in the request message

(I think there's a more explicit reference, let me see whether I can dig it up)

vchudnov-g: The internal grpc transcoding doc mentions encoding the leaf fields

PR #770, opened 2021-06-24
WIP leaf fields

software-dov:
Companion PR to https://github.com/googleapis/gapic-generator-go/pull/676


client/compliance_client.go
```diff
@@ -585,6 +654,894 @@ func (c *complianceGRPCClient) CancelOperation(ctx context.Context, req *longrun
...
+	if req.GetInfo().GetFChild().GetPChild().GetFDouble() != 0 {
+		params.Add("info.fChild.pChild.fDouble", fmt.Sprintf("%v", req.GetInfo().GetFChild().GetPChild().GetFDouble()))
+	}
+	if req.GetInfo().GetFChild().GetPChild().GetFString() != "" {
+		params.Add("info.fChild.pChild.fString", fmt.Sprintf("%v", req.GetInfo().GetFChild().GetPChild().GetFString()))
+	}
+	if req.GetInfo().GetFChild().GetPContinent() != 0 {
+		params.Add("info.fChild.pContinent", fmt.Sprintf("%v", req.GetInfo().GetFChild().GetPContinent()))
+	}
+	if req.GetInfo().GetFChild().GetPDouble() != 0 {
```
vchudnov-g, line 813: For the `P*` fields, which are proto3 `optional`, we want to include them if they have been set, even if they've been set to the zero value. So you probably want `....GetFChild().PDouble != nil`, or better yet, calling a method like `....GetFChild().HasPDouble()` if such a method exists.

client/compliance_client.go
```diff
@@ -585,6 +654,894 @@ func (c *complianceGRPCClient) CancelOperation(ctx context.Context, req *longrun
...
+	jsonReq, err := m.Marshal(req)
+	if err != nil {
+		return nil, err
+	}
+
+	baseUrl, _ := url.Parse(c.endpoint)
+	baseUrl.Path += fmt.Sprintf("/v1beta1/repeat:query")
+
+	params := url.Values{}
+	if req.GetInfo().GetFBool() {
```
vchudnov-g, line 768: I'm OK with this structure for expediency, but I think long term it might be desirable to use proto3 reflection to construct the list of query params. That will make the code more readable, and will allow us to handle recursive message definitions (which is both elegant and, I believe, allowed)

client/compliance_client.go
```diff
@@ -585,6 +654,894 @@ func (c *complianceGRPCClient) CancelOperation(ctx context.Context, req *longrun
...
+	if req.GetInfo().GetFChild().GetPBool() {
+		params.Add("info.fChild.pBool", fmt.Sprintf("%v", req.GetInfo().GetFChild().GetPBool()))
+	}
+	if req.GetInfo().GetFChild().GetPChild().GetFBool() {
+		params.Add("info.fChild.pChild.fBool", fmt.Sprintf("%v", req.GetInfo().GetFChild().GetPChild().GetFBool()))
+	}
+	if req.GetInfo().GetFChild().GetPChild().GetFDouble() != 0 {
+		params.Add("info.fChild.pChild.fDouble", fmt.Sprintf("%v", req.GetInfo().GetFChild().GetPChild().GetFDouble()))
+	}
+	if req.GetInfo().GetFChild().GetPChild().GetFString() != "" {
```
vchudnov-g, line 807: Note that if `PChild` (which is `optional`) has not been set at all, we don't need to dive into the fields of  the empty message returned by by `GetPChild()`. And also, this could cause problems or unwieldiness for recursive definitions. See my other comments.

client/compliance_client.go
vchudnov-g: The Showcase server is doing the opposite (json to proto) in a couple of places, like [this](https://github.com/googleapis/gapic-showcase/blob/04cab450b3ad6b7519ec7754db40fe633d7f47a6/util/genrest/resttools/populatefield.go#L63). That may be helpful for you. In particular, the [protoreflect](https://pkg.go.dev/google.golang.org/protobuf/reflect/protoreflect) package.

client/compliance_client.go
software-dov: Damn, you're right. I completely neglected that possibility.

software-dov: The more I think about this the more I'm in favor of not handling this case specially. It looks like go does not provide any mechanism for determining whether an optional field has been set.

Consider the following:
```proto
syntax = "proto3";
package test.thing;

option go_package = "server/genproto";

message Inner {}

message Outer {
        Inner required_inner = 1;
        optional Inner optional_inner = 2;
        int32 required_int = 3;
        optional int32 optional_int = 4;
        oneof variant {
              Inner inner_variant = 5;
              int32 int32_variant = 6;
        }
}
```

This gives us the following:
```go
type Outer struct {
// ...
	RequiredInner *Inner `protobuf:"bytes,1,opt,name=required_inner,json=requiredInner,proto3" json:"required_inner,omitempty"`
	OptionalInner *Inner `protobuf:"bytes,2,opt,name=optional_inner,json=optionalInner,proto3,oneof" json:"optional_inner,omitempty"`
// ...
}

func (x *Outer) GetRequiredInner() *Inner {
	if x != nil {
		return x.RequiredInner
	}
	return nil
}

func (x *Outer) GetOptionalInner() *Inner {
	if x != nil {
		return x.OptionalInner
	}
	return nil
}
```

That's all the support we get for `optional_inner`. There does not seem to be a `HasOptionalInner` method.

Recursive messages are already handled by logic in `getLeafs`, so I don't think that's a huge problem.

client/compliance_client.go
software-dov: Such methods do not exist, unfortunately. For optional leaf primitive fields I think we are forced to do the double check of nil and and value.

client/compliance_client.go
noahdietz: This is one of those moments where it is OK to directly reference the field. `if x.OptionalInner != nil`.

Setting & checking presence is an acceptable use of direct reference. Retrieving a value should use the accessor, as best practice.

client/compliance_client.go
software-dov: That's extremely helpful. Thanks!

YoutacVA (approved):

YoutacVA (approved):

YoutacVA (approved):

PR #802, opened 2021-07-22
feat: add Compliance RPCs that transcode to PUT and PATCH HTTP verbs.

vchudnov-g:
These methods are now exercised in `compliance_suite.json`.

Includes tests.

### DIFF
--- a/client/gapic_metadata.json
+++ b/client/gapic_metadata.json
@@ -55,6 +55,16 @@
                 "RepeatDataBodyInfo"
               ]
             },
+            "RepeatDataBodyPatch": {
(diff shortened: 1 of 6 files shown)


client/gapic_metadata.json
```diff
@@ -55,6 +55,16 @@
                 "RepeatDataBodyInfo"
               ]
             },
+            "RepeatDataBodyPatch": {
```
noahdietz, line 58: Nit: this will be updated in the following regen PR, but its fine to keep here too.

noahdietz (approved):

PR #806, opened 2021-07-23
feat: implement non-standard pagination max_results in Echo service

vchudnov-g:
Includes tests


schema/google/showcase/v1beta1/echo.proto
```diff
@@ -163,6 +173,20 @@ message PagedExpandRequest {
...
 }
 
+// The request for the PagedExpandLegacy method. New APIs should NOT
+// use this pattern; use PagedExpandRequest instead.
+message PagedExpandLegacyRequest {
+  // The string to expand.
+  string content = 1 [(google.api.field_behavior) = REQUIRED];
+
+  // The amount of words to returned in each page.  This is a legacy, non-standard pattern that
+  // violates http://https://google.aip.dev/158
```
noahdietz, line 183: Nit: can you wrap the reasoning/commentary in an aip.dev/not-precedent comment please?

noahdietz (approved):
LGTM. Will this RPC be "expanded" (no pun intended) to include the map paging once supported? Or will that be a new RPC?

schema/google/showcase/v1beta1/echo.proto
vchudnov-g: Good idea! I forgot about that.

vchudnov-g:
My plan is to have a separate RPC for paged responses that are maps.

PR #813, opened 2021-07-28
Companion PR for paged maps

software-dov:


schema/google/showcase/v1beta1/echo.proto
```diff
@@ -115,8 +115,47 @@ service Echo {
       body: "*"
     };
   };
+    rpc MapExpand(MapExpandRequest) returns (MapExpandResponse) {
```
vchudnov-g, line 118: FYI: The code I have in progress will add these mapped-response paginated methods (under a different name). If you want, I could share that proto with you now (I'm working on unit tests, and will need your changes released first to submit the PR and release)

vchudnov-g: Here's an older version of my work with paginated maps. You can get the protos and Showcase server code from it: https://github.com/vchudnov-g/gapic-showcase/tree/regapic-pagination

software-dov: That's perfect. I'll not submit these changes since they add the dummy MapExpand methods but keep this PR as is for ease of review.

PR #818, opened 2021-07-31
feat: Implement server streaming over REST

vchudnov-g:
This returns all the streamed server responses as one single HTTP response, containing a JSON list with all the individual responses.

```
❯ curl -i -X POST http://localhost:7469/v1beta1/echo:expand -d'{"content":"it was a dark and stormy night"}' -H "Con
tent-Type: application/json" -H "X-Goog-Api-Client: rest/0.0.0 gapic/0.0.0"
HTTP/1.1 200 OK
Date: Sat, 31 Jul 2021 00:23:58 GMT
Content-Length: 376

{
{
  "content": "it",
  "severity": "UNNECESSARY"
},
{
  "content": "was",
  "severity": "UNNECESSARY"
},
{
  "content": "a",
  "severity": "UNNECESSARY"
},
{
  "content": "dark",
  "severity": "UNNECESSARY"
},
{
  "content": "and",
  "severity": "UNNECESSARY"
},
{
  "content": "stormy",
  "severity": "UNNECESSARY"
},
{
  "content": "night",
  "severity": "UNNECESSARY"
}
}      
```

This is a DRAFT for now because it does not include tests, but feel free to patch for local use. I am thinking about the best way to implement tests.


util/genrest/testdata/TestConstructServerStreamer.go.baseline
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// ListJSON returns a list of all the accumulated responses, in JSON format.
+func (streamer *Catalog_BaseServerStreamer) ListJSON() string {
+   return fmt.Sprintf("{\n%s\n}", strings.Join(streamer.responses, ",\n"))
+}
+
+func (streamer *Catalog_BaseServerStreamer) initialize() {
+   streamer.marshaler = resttools.ToJSON()
+}
```
software-dov, line 27: I don't think it matters too much, but why have the accumulate do initialize once/idemontently, instead of creating the streamer via a factory function and returning it with the correct marshaler?

util/genrest/goviewcreator.go
```diff
@@ -270,6 +298,72 @@ func NewView(model *gomodel.Model) (*goview.View, error) {
...
+	baseHelper.P(`func (streamer *%s) ListJSON() string {`, baseStreamerType)
+	baseHelper.P(`   return fmt.Sprintf("{\n%%s\n}", strings.Join(streamer.responses, ",\n"))`)
+	baseHelper.P(`}`)
+	baseHelper.P(``)
+	baseHelper.P(`func (streamer *%s) initialize() {`, baseStreamerType)
+	baseHelper.P(`   streamer.marshaler = resttools.ToJSON()`)
+	baseHelper.P(`}`)
+	baseHelper.P(``)
+	baseHelper.P(`func (streamer *%s) Context() context.Context {`, baseStreamerType)
+	baseHelper.P(`   return context.Background()`)
```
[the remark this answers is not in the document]
vchudnov-g: It's not needed for compilation as I am embedding the interface that requires it.....but it is required at runtime because this method does get called. (I just tried it, and it does crash without it!)

util/genrest/testdata/TestConstructServerStreamer.go.baseline
vchudnov-g: No strong reason; we could do it either way.

In some sense it seems cleaner and more concise this way, because otherwise we would have one  factory function for `Echo_ExpandServer` (and other streaming RPCs) , and each of those would either call the factory function for  `Echo_BaseServerStreamer` or explicitly initialize its internals. This way it's more self-contained , and `accumulate()` enforces its prerequisites being satisfied.

vchudnov-g (commented):
I am coming back to this PR after a long time away. I rebased this onto the current `HEAD` to get the latest changes; hope this doen't complicate your review.

In a subsequent PR I'll deal with actually enforcing chunked encoding with time delays, which is not happening yet (run the `curl` command in the PR description with the `--raw` flag).

summer-ji-eng (approved):
LGTM, expecting the change with transfer-encoding chunked data. 🚀

PR #877, opened 2021-09-16
chore: fix code-suggester primary branch

noahdietz:


vchudnov-g:
When we switch from `master` to `main`, we'll have to remember to change this instance as well. I assume we'll need to grep for `master` anyway....

vchudnov-g (approved):

PR #930, opened 2021-11-24
fix!: add `parent` to method signature for `Messaging.SearchBlurbs()`

vchudnov-g:
BREAKING CHANGE: this changes the flattened method signature for the Messaging.SearchBlurbs() RPC to require  a `parent` parameter.

Without this change, the flattened method does not work, as the required field parent is not supplied.


vam-google (approved):
LGTM

PR #970, opened 2022-01-14
feat: add ability to echo headers & sample routing annotations on echo method

alicejli:
As part of supporting dynamic routing annotations per `google/api/routing.proto`, this adds the following:

1) Several sample routing annotations to the `Echo.Echo` method
2) The ability for each `Echo` service to echo the contents set with the `x-goog-request-params` metadata key to the response headers

An example of how to set and read the headers in Go:

```
func main() {
	conn, err := grpc.Dial("localhost:7469", grpc.WithInsecure())
	// This is what the generated client library should be creating
        md := metadata.New(map[string]string{"x-goog-request-params": "testHeader"})
	if err != nil {
		log.Fatal(err)
	}
	defer conn.Close()
	opt := option.WithGRPCConn(conn)
	ctx := metadata.NewOutgoingContext(context.Background(), md)

	echo, err := showcase.NewEchoClient(ctx, opt)
	if err != nil {
		log.Fatal(err)
	}
	defer echo.Close()

	content := "HELLO WORLD"

	req := &showcasepb.EchoRequest{
		Response: &showcasepb.EchoRequest_Content{
			Content: content,
		},
	}
	// Very important step! Another metadata instance must be provided to the `grpc.Header()` with which to capture them.
	mdForHeaders := metadata.New(map[string]string{})

	resp, err := echo.Echo(ctx, req, gax.WithGRPCOptions(grpc.Header(&mdForHeaders)))

	fmt.Printf("This is the response %v\nThis is the header:%v", resp, mdForHeaders.Get("x-goog-request-params")

	if err != nil {
		log.Fatal(err)
	}
}
```

The result should be:
```
This is the response content:"HELLO WORLD"
This is the header:[testHeader]
```

Each language microgenerator can use the following sample tests to the `Echo` method to confirm the dynamic routing annotations are being generated properly.

1) EchoRequest without `header` or `other_header` field specified
Header should be empty.

2) 
```
EchoRequest{
   other_header: “waldo”
}
``` 
Expected Header: `baz=waldo`

3)
```
EchoRequest{
   header: “foo”
}
``` 
Expected Header: `routing_id=foo&header=foo`

4) 
```
EchoRequest{
   header: “projects/foo”
}
``` 
Expected Header: `routing_id=projects/foo&header=projects/foo`

5) 
```
EchoRequest{
   header: “projects/foo/instances/quuz/rows”
}
``` 
Expected Header: `instance_id=instances/quuz&table_name=projects/foo/instances/quuz/rows&header&super_id=projects/foo&routing_id=projects/foo/instances/quuz/rows&header=projects/foo/instances/quuz/rows`

6) 
```
EchoRequest{
   header: “regions/foo/zones/bar/rows””
}
``` 
Expected Header: `table_name=regions/foo/zones/bar/rows&routing_id=regions/foo/zones/bar/rows&header=regions/foo/zones/bar/rows`

7) 
```
EchoRequest{
   header: “projects/foo/instances/bar”
}
``` 
Expected Header: `super_id=projects/foo&routing_id=projects/foo/instances/bar&header=projects/foo/instances/bar`

8) 
```
EchoRequest{
   header: “thud"
   other_header: “projects/foo/instances/bar”
}
``` 
Expected Header: `quz=projects/foo&baz=projects/foo/instances/bar&routing_id=thud&header=thud`

### DIFF
--- a/server/services/echo_service.go
+++ b/server/services/echo_service.go
@@ -17,6 +17,7 @@ package services
 import (
 	"context"
 	"io"
+	"log"
 	"strconv"
 	"strings"
 	"time"
@@ -46,21 +47,26 @@ func (s *echoServerImpl) Echo(ctx context.Context, in *pb.EchoRequest) (*pb.Echo
 	if err != nil {
 		return nil, err
 	}
+	echoHeaders(ctx)
 	echoTrailers(ctx)
 	return &pb.EchoResponse{Content: in.GetContent(), Severity: in.GetSeverity()}, nil
 }
 
 func (s *echoServerImpl) Expand(in *pb.ExpandRequest, stream pb.Echo_ExpandServer) error {
 	for _, word := range strings.Fields(in.GetContent()) {
+		log.Printf("range_stream: %v:", stream)
 		err := stream.Send(&pb.EchoResponse{Content: word})
 		if err != nil {
+			log.Printf("err_stream: %v:", stream)
 			return err
 		}
 	}
+	echoStreamingHeaders(stream)
 	if in.GetError() != nil {
 		return status.ErrorProto(in.GetError())
 	}
 	echoStreamingTrailers(stream)
+	log.Printf("ending_stream: %v:", stream)
 	return nil
 }
 
@@ -70,6 +76,7 @@ func (s *echoServerImpl) Collect(stream pb.Echo_CollectServer) error {
 	for {
 		req, err := stream.Recv()
 		if err == io.EOF {
+			echoStreamingHeaders(stream)
 			echoStreamingTrailers(stream)
 			return stream.SendAndClose(&pb.EchoResponse{Content: strings.Join(resp, " ")})
 		}
@@ -101,6 +108,7 @@ func (s *echoServerImpl) Chat(stream pb.Echo_ChatServer) error {
 		if s != nil {
 			return s
 		}
+		echoStreamingHeaders(stream)
 		stream.Send(&pb.EchoResponse{Content: req.GetContent()})
 	}
 }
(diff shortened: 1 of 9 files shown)


server/services/echo_service.go
```diff
@@ -46,21 +47,26 @@ func (s *echoServerImpl) Echo(ctx context.Context, in *pb.EchoRequest) (*pb.Echo
...
 	echoTrailers(ctx)
 	return &pb.EchoResponse{Content: in.GetContent(), Severity: in.GetSeverity()}, nil
 }
 
 func (s *echoServerImpl) Expand(in *pb.ExpandRequest, stream pb.Echo_ExpandServer) error {
 	for _, word := range strings.Fields(in.GetContent()) {
+		log.Printf("range_stream: %v:", stream)
 		err := stream.Send(&pb.EchoResponse{Content: word})
 		if err != nil {
+			log.Printf("err_stream: %v:", stream)
```
noahdietz, line 60: Please remove these log statements.

server/services/echo_service.go
```diff
@@ -46,21 +47,26 @@ func (s *echoServerImpl) Echo(ctx context.Context, in *pb.EchoRequest) (*pb.Echo
...
 			return err
 		}
 	}
+	echoStreamingHeaders(stream)
 	if in.GetError() != nil {
 		return status.ErrorProto(in.GetError())
 	}
 	echoStreamingTrailers(stream)
+	log.Printf("ending_stream: %v:", stream)
```
noahdietz, line 69: Remove this too

noahdietz (commented):
Looks great.

server/services/echo_service.go
alicejli: Oops, thanks for catching. Removed!

server/services/echo_service.go
alicejli: Done! Thanks again for catching and reviewing.

noahdietz (approved):
I'm happy with this. I will let you, @alicejli decide if you want eyes from the other reviewers.

PR #980, opened 2022-01-31
fix: update genproto with new echo

alicejli:
Somehow in https://github.com/googleapis/gapic-showcase/pull/970, the updated `echo.pb.go` and related files did not make it in. This updates the Echo proto to take into account additional fields for testing headers.


noahdietz:
I don't think we need this PR anymore because I merged #962.

I forgot to mention that generated code changes are made by that regen job.

alicejli:
Whoops thank you @noahdietz ; turns out I hadn't updated my repo since that PR got merged. Will close this out; thank you!

PR #1082, opened 2022-05-06
feat(genrest): format rest errors as Google Errors

noahdietz:
This modifies the REST server generator to wrap errors in accordance with [Google Cloud's API Error format](https://cloud.google.com/apis/design/errors). Furthermore, it unpacks the gRPC status-based errors returned by the gRPC server handlers, and converts the gRPC Code into the proper HTTP status code.

Tested locally via Go generator REGAPIC integration tests.

### DIFF
--- a/util/genrest/resttools/error_response.go
+++ b/util/genrest/resttools/error_response.go
@@ -1,0 +1,64 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package resttools
+
+import (
+	"encoding/json"
+	"net/http"
+
+	"google.golang.org/api/googleapi"
+	"google.golang.org/grpc/codes"
+)
+
+// Derived from internal source: go/http-canonical-mapping.
+var GRPCToHTTP map[codes.Code]int = map[codes.Code]int{
+	codes.OK:                 http.StatusOK,
+	codes.Canceled:           499, // There isn't a Go constant ClientClosedConnection
+	codes.Unknown:            http.StatusInternalServerError,
+	codes.InvalidArgument:    http.StatusBadRequest,
+	codes.DeadlineExceeded:   http.StatusGatewayTimeout,
+	codes.NotFound:           http.StatusNotFound,
+	codes.AlreadyExists:      http.StatusConflict,
+	codes.PermissionDenied:   http.StatusForbidden,
+	codes.Unauthenticated:    http.StatusUnauthorized,
+	codes.ResourceExhausted:  http.StatusTooManyRequests,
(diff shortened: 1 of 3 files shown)


util/genrest/resttools/error_response.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// Google API Errors, as defined by https://cloud.google.com/apis/design/errors
+// will consist of a googleapi.Error nested as the key `error` in a JSON object.
+// So we must create such a structure to wrap our googleapi.Error in.
+type googleAPIError struct {
+	Error *googleapi.Error `json:"error"`
+}
+
+func ErrorResponse(w http.ResponseWriter, status int, message string, details ...interface{}) {
```
vchudnov-g, line 53: I might s/`status`/`httpStatus`/ and/or mention in the func godoc that this is the HTTP status code for the error.

vchudnov-g, line 53: Godoc for the func?

util/genrest/resttools/error_response_test.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+		err := googleapi.CheckResponse(got.Result())
+		var gerr *googleapi.Error
+		if !errors.As(err, &gerr) {
+			t.Fatalf("%s: Expected response to be a googleapi.Error, but got %v", tst.name, err)
+		}
+		if diff := cmp.Diff(gerr.Message, tst.message); diff != "" {
+			t.Errorf("%s: got(-),want(+):%s\n", tst.name, diff)
+		}
+		if diff := cmp.Diff(gerr.Details, tst.details); diff != "" {
```
vchudnov-g, line 57: Do you want to add `details` to some of your test cases?

util/genrest/resttools/error_response.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"encoding/json"
+	"net/http"
+
+	"google.golang.org/api/googleapi"
+	"google.golang.org/grpc/codes"
+)
+
+// Derived from internal source: go/http-canonical-mapping.
+var GRPCToHTTP map[codes.Code]int = map[codes.Code]int{
```
vchudnov-g, line 26: nit: start doc with name of element being described.

vchudnov-g, line 26: A nit on principle, unlikely to come up in practice:

Would it be better to have the table be an internal variable, and have a `func GRPCToHttp(code codes.Code) int` so that if there isn't a key with the value of `code`, we treat is as `codes.Unknown`?

vchudnov-g (commented):
Thanks for doing this!

vchudnov-g (approved):
Feel free to submit once you've thought about my suggestions.

util/genrest/resttools/error_response.go
noahdietz: Yeah great idea! I'll add that.

util/genrest/resttools/error_response.go
noahdietz: Sure, done!

util/genrest/resttools/error_response_test.go
noahdietz: Sure, I used a simple type, not a protobuf type, as I don't really want to test the implementation of serializing the googleapi.Error into JSON, but we can circle back later if we want to add google.rpc.ErrorInfo, for example.

util/genrest/resttools/error_response.go
noahdietz: Done!

vchudnov-g (approved):
Looks great! Thanks!

PR #1118, opened 2022-06-10
feat: Support LRO mixins over REST

vchudnov-g:
For future reference: Some of this code was adapted from gapic-generator-go. 

https://github.com/googleapis/gapic-generator-go/blob/c9e3ce74af160bf72a8d140f2d4ecb167d35df25/internal/gengapic/mixins.go#L66

https://github.com/googleapis/gapic-generator-go/blob/f40c8300f39be4584ff195f1cbc0488bebd563ae/internal/gengapic/generator.go#L103-L117


vchudnov-g:
Note that we also need to  add an http rule to the WaitOperation rpc in api-common-proto submodule

```
  option (google.api.http) = {
      get: "/v1/{name=operations/**}:wait"
    };
```

What's SOP for doing this?

vchudnov-g:
Sent out a PR with the HTTP rule: https://github.com/googleapis/api-common-protos/pull/111

vchudnov-g:
I took the approach of generating all the methods in the mixin protos. This means that if they are not referenced in the service config, their http bindings do not get overriden.

And adding `WaitOperation` to this service config causes a conflict with `EchoClient.WaitOperation`

server/genrest/genrest.go
```diff
@@ -40,6 +40,11 @@ func RegisterHandlers(router *gmux.Router, backend *services.Backend) {
 	router.HandleFunc("/v1beta1/repeat/{info.fString:first/.+}/{info.fChild.fString:second/.+}:pathtrailingresource", rest.HandleRepeatDataPathTrailingResource).Methods("GET")
 	router.HandleFunc("/v1beta1/repeat:bodyput", rest.HandleRepeatDataBodyPut).Methods("PUT")
 	router.HandleFunc("/v1beta1/repeat:bodypatch", rest.HandleRepeatDataBodyPatch).Methods("PATCH")
+	router.HandleFunc("/v1beta1/operations", rest.HandleListOperations).Methods("GET")
+	router.HandleFunc("/v1beta1/{name:operations/.+}", rest.HandleGetOperation).Methods("GET")
+	router.HandleFunc("/v1beta1/{name:operations/.+}:delete", rest.HandleDeleteOperation).Methods("DELETE")
+	router.HandleFunc("/v1beta1/{name:operations/.+}:cancel", rest.HandleCancelOperation).Methods("POST")
+	router.HandleFunc("/v1/{name:operations/.+}:wait", rest.HandleWaitOperation).Methods("GET")
```
vchudnov-g, line 47: I took the approach of generating all the methods in the mixin protos. This means that if they are not referenced in the service config, their http bindings do not get overriden.

And adding `WaitOperation` to this service config causes a conflict with `EchoClient.WaitOperation`

vchudnov-g: So we either live with the extra methods, or implement a strategy closer to that in the Go generator to ensure we don't generate the extra methods. Thoughts?

vchudnov-g:
Restructured the code so we only generate code for the listed mixin methods.

Service config path is still hard-coded at the moment.

go.mod
```diff
@@ -13,6 +14,7 @@ require (
 	github.com/spf13/viper v1.12.0
 	golang.org/x/oauth2 v0.0.0-20220608161450-d0670ef3b1eb
 	golang.org/x/sync v0.0.0-20220601150217-0de741cfad7f
+	golang.org/x/xerrors v0.0.0-20220517211312-f3a8303e98df
```
noahdietz, line 17: Note to self: I don't think we need this, we can use stdlib `errors` because we support go1.13 <

schema/google/showcase/v1beta1/showcase_v1beta1.yaml
```diff
@@ -78,8 +78,8 @@ http:
   - selector: google.longrunning.Operations.ListOperations
     get: '/v1beta1/operations'
   - selector: google.longrunning.Operations.GetOperation
-    get: '/v1beta1/{name=/operations/*}'
+    get: '/v1beta1/{name=operations/**}'
   - selector: google.longrunning.Operations.DeleteOperation
-    delete: '/v1beta1/{name=/operations/*}'
+    delete: '/v1beta1/{name=operations/**}:delete'
```
noahdietz, line 83: Note to self: we shouldn't need the `:delete` suffix, this isn't a custom method.

util/genrest/protomodelcreator.go
```diff
@@ -141,3 +167,108 @@ func NewRESTRequestPattern(rule *annotations.HttpRule) (*protomodel.RESTRequestP
...
+		}
+	}
+	return files
+}
+
+var mixinDescriptors map[string][]*descriptor.FileDescriptorProto
+
+func init() {
+	mixinDescriptors = map[string][]*descriptor.FileDescriptorProto{
+		"google.longrunning.Operations": {
```
noahdietz, line 270: Note to self: add the other mixin services too (Locations, IAMPolicy)

noahdietz (approved):

schema/google/showcase/v1beta1/showcase_v1beta1.yaml
vchudnov-g: True. Thanks for the catch. (And fix)

util/genrest/protomodelcreator.go
vchudnov-g: I was getting some (probably trivial) errors with Locations, so I just excluded the others for now so we could get LRO working. But as you see, I copied your structure from gapic-generator-go so the others would be easy to add back in.

PR #1150, opened 2022-07-21
feat: add binding testing and multiple bindings test data to compliance service

viacheslav-rostovtsev:
Split into semantically meaningful commits for reading convenience.


schema/google/showcase/v1beta1/compliance.proto
```diff
@@ -116,10 +119,13 @@ message RepeatRequest {
   optional int32 p_int32 = 7;
   optional int64 p_int64 = 8;
   optional double p_double = 9;
+
+  optional string intended_binding_uri = 10;
```
vchudnov-g, line 123: I would place this near the `server_verify` field, since it's not testing the data fromat preservation like the rest of the compliance suite is

util/genrest/resttools/constants.go
```diff
@@ -56,3 +56,6 @@ func init() {
 
 	RegexLiteral = fmt.Sprintf(`^[%s]+`, CharsLiteral)
 }
+
+// A key-type for storing binding URI value in the Context
+type BindingURIKey string
```
vchudnov-g, line 61: all the uses of this are `BindingUriKey("bindingUri")`, AFICT. Why not define `const BindingUriKey string="bindingUri"`?

vchudnov-g (commented):
Did a first pass, but let me do a second pass next week. I think it's looking good, but I want to think some things through.

cmd/gapic-showcase/endpoint_test.go
```diff
@@ -174,7 +175,7 @@ func TestRESTCalls(t *testing.T) {
 			log.Fatal(err)
 		}
 		if got, want := string(body), testCase.want; noSpace(got) != noSpace(want) {
-			t.Errorf("testcase %2d: body: got `%s`, want %s", idx, got, want)
+			t.Errorf("testcase %2d: body: got `%s`, want `%s`", idx, noSpace(got), noSpace(want))
```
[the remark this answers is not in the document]
viacheslav-rostovtsev: :+1: 
I think being able to see the real assert condition is more convenient. In my case e.g. there was an issue with a tab, invisible in the 'readable' version.

schema/google/showcase/v1beta1/compliance.proto
viacheslav-rostovtsev: :+1:

util/genrest/resttools/constants.go
viacheslav-rostovtsev: :+1:

viacheslav-rostovtsev: the constant has it's own string-derived type instead of string because (as I understand) the Context is a shared space so if someone else wants to have a "bindingUri" it'll be in a separate type.

cmd/gapic-showcase/compliance_suite_test.go
```diff
@@ -217,6 +217,16 @@ func prepRepeatDataPathResourceTest(request *genproto.RepeatRequest) (verb strin
 	name = "Compliance.RepeatDataPathResource"
 	info := request.GetInfo()
 
+	if strings.HasPrefix(info.GetFString(), "first/") {
+		return prepRepeatDataPathResourceTestFirstBinding(request)
```
vchudnov-g, line 221: (no action needed) I think this is fine, but it makes me think in the future we should change the `first/` and `second/` in the URL templates to other names for clarity, like `blue` and `cyan` or something.....

server/services/compliance_service.go
```diff
@@ -51,6 +53,20 @@ func (csi *complianceServerImpl) requestMatchesExpectation(received *pb.RepeatRe
...
+		intendedBinding := expectedRequest.GetIntendedBindingUri()
+		if intendedBinding != binding {
+			return fmt.Errorf("(ComplianceSuiteWrongBindingError) request %q was transcoded to the wrong binding (expected: %s; actual %s)", name, intendedBinding, binding)
+		}
+	}
+
+	// Separately checking that the binding in the client request matches binding in the test suite.
+	// This guards against the test suite file being wrong on the client.
+	if expectedRequest.GetIntendedBindingUri() != received.GetIntendedBindingUri() {
+		return fmt.Errorf("(ComplianceSuiteRequestBindingMismatchError) intended binding of request %q do not match test suite (expected: %s, received: %s)", name, expectedRequest.GetIntendedBindingUri(), received.GetIntendedBindingUri())
```
vchudnov-g, line 67: nit: s/do not/does not/

util/genrest/resttools/constants.go
vchudnov-g: Oh, I see what you mean about the type. The [doc](https://pkg.go.dev/context#WithValue) stipulates using non-built-in types to avoid collisions with other packages, since the type, not just the name,  presumably plays a part in the lookup.

cmd/gapic-showcase/endpoint_test.go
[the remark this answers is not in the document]
vchudnov-g: Ah, so the problem might be that `noSpace` does not remove `\t`. Maybe it should, and then you can have the asserts use `noSpace` but the `t.Errorf` remain readable.

vchudnov-g (approved):

PR #1151, opened 2022-07-22
fix: remove broken test cases in compliance

viacheslav-rostovtsev:
Removing RPCs that should not be able to transcode the request given in the test suite.
The test gives the following value for the `info.fString` field: `"first/hello/second/greetings"`.

In their respective single http binding uri templates:
- `RepeatDataSimplePath` RPC has a binding `{info.f_string}` which is a shorthand for `{info.f_string=*}`, the pattern `*` does not allow slashes and will not match the value.
- `RepeatDataPathResource` and RepeatDataPathTrailingResource RPCs have the same binding `{info.f_string=first/*}` which does not allow more slashes after the `first/` part and will not match the value.

### DIFF
--- a/server/services/compliance_suite.json
+++ b/server/services/compliance_suite.json
@@ -145,7 +145,7 @@
     },
     {
       "name": "Fully working conversions, resources",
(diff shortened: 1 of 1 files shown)


vchudnov-g (approved):
Oh, good catch! Thanks for fixing this. I'll take an AI to figure out why this wasn't failing in Showcase or in Go; we probably weren't executing this test case, which is a bit troubling.

PR #1159, opened 2022-07-27
feat(regapic): accept numeric enums, allow testing enum round trips

vchudnov-g:
This introduces the RPCs `Compliance.GetEnum()` and `Compliance.VerifyEnum()`. The first returns one enum value that is either known or unknown to the client. The client can then send that response as the request message to the second RPC, which will verify that the value received was the same as the one originally sent. This thus tests that enum fields that are populated by the server with an unknown-to-the-client value (maybe because the client hasn't been regenerated for the latest proto version) and which are untouched by client code can be sent back to the server while preserving their value. This is particularly important in the REST transport case, and is the reason why we are separately implementing a feature to request that the server, if using REST, send all enums as numbers rather than strings.                   
    
To test various types of responses from the command line, you can use the following commands:
    
```
curl -i  -H "Content-Type: application/json" -H "X-Goog-Api-Client: rest/0.0.0 gapic/0.0.0" http://localhost:7469/v1beta1/compliance/enum\?unknownEnum\=true
    
curl -i  -H "Content-Type: application/json" -H "X-Goog-Api-Client: rest/0.0.0 gapic/0.0.0" http://localhost:7469/v1beta1/compliance/enum\?unknownEnum\=false
    
curl -i -X POST  -H "Content-Type: application/json" -H "X-Goog-Api-Client: rest/0.0.0 gapic/0.0.0" http://localhost:7469/v1beta1/compliance/enum?request.unknownEnum=true&continent=57                                                                                                                                                  
```
    
To test requesting and verifying an enum from the command line:
    
```
echo -e "\n**** Request enum:  ****\n" && \
  REGAPIC_ENUM=$(curl -i  -H "Content-Type: application/json" -H "X-Goog-Api-Client: rest/0.0.0 gapic/0.0.0" http://localhost:7469/v1beta1/compliance/enum\?unknownEnum\=true | tee /dev/tty | grep continent | grep -o -E -- '-?[0-9]+')  && \                                                                                    
  echo -e "\n\n**** Verify enum:  ****\n" && \
  curl -i -X POST  -H "Content-Type: application/json" -H "X-Goog-Api-Client: rest/0.0.0 gapic/0.0.0" http://localhost:7469/v1beta1/compliance/enum?request.unknownEnum=true&continent=$REGAPIC_ENUM                                                                                                                               
```

(GAPICs should simply call the corresponding generated methods analogously.)

As part of enabling this functionality, Showcase now allows receiving REST enum values JSON-encoded as numbers rather than strings. Earlier, there was an explicit prohibition against this.


util/genrest/resttools/populatefield.go
```diff
@@ -152,10 +152,16 @@ func PopulateOneField(protoMessage proto.Message, fieldPath string, fieldValues
...
 		case protoreflect.EnumKind:
-			if _, parseError = strconv.ParseFloat(value, 32); parseError == nil {
-				return fmt.Errorf("(EnumValueNotStringError) enum value %q for field path %q appears to be a number rather than an identifier", fieldName, fieldPath)
+			var parsedValue protoreflect.EnumNumber
+			if numericValue, err := strconv.ParseFloat(value, 32); err == nil {
+				parsedValue = protoreflect.EnumNumber(numericValue)
+			} else {
+				enum := fieldDescriptor.Enum().Values().ByName(protoreflect.Name(value))
+				if enum == nil {
+					return fmt.Errorf("(UnknownEnumStringError) unknown enum symbol %q for field path %q", value, fieldPath) // FIXME(vchudnov): add tests
```
noahdietz, line 161: Is the FIXME still relevant?

vchudnov-g: Oops, no. I added the test. Nice catch!

noahdietz:
Ping me when the tests are passing so I can review the end state :)

vchudnov-g:
OK, tests passed. I was working off an older `HEAD`.

noahdietz (approved):

PR #1187, opened 2022-08-26
fix: Update README with latest version number

vchudnov-g:
Also do some copyediting fixes.


vchudnov-g:
I thought we had a script that would update all the version references throughout the repo. We should include this reference in there.

@noahdietz

noahdietz (approved):

PR #1198, opened 2022-08-31
fix!: use resource field in http body for Updates

noahdietz:
Fixes the `google.api.http` `body` fields for UpdateUser, UpdateBlurb, and UpdateRoom to use the field in the request representing the resource, instead of `*`.

This also adds support for parsing some of the well known protobuf types that may be present as JSON encoded query parameters and that have special JSON encoding formats (see https://developers.google.com/protocol-buffers/docs/proto3#json). This API only makes use of Timestamp, Duration, and FieldMask.

### DIFF
--- a/schema/google/showcase/v1beta1/identity.proto
+++ b/schema/google/showcase/v1beta1/identity.proto
@@ -58,7 +58,7 @@ service Identity {
   rpc UpdateUser(UpdateUserRequest) returns (User) {
     option (google.api.http) = {
       patch: "/v1beta1/{user.name=users/*}"
-      body: "*"
+      body: "user"
     };
   }
 
--- a/util/genrest/resttools/populatefield.go
+++ b/util/genrest/resttools/populatefield.go
@@ -19,9 +19,30 @@ import (
 	"strconv"
 	"strings"
 
+	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/reflect/protoreflect"
 )
+
+var wellKnownTypes = map[string]bool{
+	// == The following are the only three common types used in this API ==
+	"google.protobuf.FieldMask": true,
+	"google.protobuf.Timestamp": true,
+	"google.protobuf.Duration":  true,
+	// == End utilized types ==
+	"google.protobuf.DoubleValue": true,
+	"google.protobuf.FloatValue":  true,
+	"google.protobuf.Int64Value":  true,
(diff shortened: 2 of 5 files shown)


schema/google/showcase/v1beta1/identity.proto
```diff
@@ -58,7 +58,7 @@ service Identity {
   rpc UpdateUser(UpdateUserRequest) returns (User) {
     option (google.api.http) = {
       patch: "/v1beta1/{user.name=users/*}"
-      body: "*"
+      body: "user"
```
vchudnov-g, line 61: Why not have `*` so that the user can specify which fields to PATCH? It seems like this is one of the prime uses cases for field masks.

util/genrest/resttools/populatefield.go
```diff
@@ -214,3 +241,22 @@ func findProtoFieldByJSONName(fieldName string, fields protoreflect.FieldDescrip
 	}
 	return result, nil
 }
+
+func parseWellKnownType(message protoreflect.Message, fieldDescriptor protoreflect.FieldDescriptor, value string) (*protoreflect.Value, error) {
+	messageFieldTypes := message.Type().(protoreflect.MessageFieldTypes)
```
vchudnov-g, line 246: For safety/future-proofing, I would add a second return param `ok` here to prevent a panic, and return an error if `!ok`

vchudnov-g (commented):
I don't understand why we have to change `body` to NOT be `*`. Could you explain? Everything else lgtm.

schema/google/showcase/v1beta1/identity.proto
noahdietz: I'm not sure what you mean. The update_mask will end up in the query params. We don't lose any functionality here, just fixing a non-compliant binding https://google.aip.dev/134

util/genrest/resttools/populatefield.go
noahdietz: Done.

schema/google/showcase/v1beta1/identity.proto
vchudnov-g: Ah, gotcha. "There must be a body key in the google.api.http annotation, and it must map to the resource field in the request message. All remaining fields should map to URI query parameters."

OK, my bad. LGTM!

vchudnov-g (approved):

noahdietz:
Ok so I am going to remove the Breaking Change indicator from the commit because I don't want to release this as a major version (at some point we really should make a v1 though).

PR #1206, opened 2022-09-06
fix: Use quotes around extreme int64 values

### LINKED ISSUE #1205: 64-bit integers should be represented as strings in JSON
compliance-suite.json includes "extreme" 64-bit integer values in numeric form. The proto3 JSON spec states that int64 values should be encoded as strings to avoid a loss of precision, although they can be parsed as numbers.

jskeet:
JSON numbers are only really expected to be precise as far as IEEE-754 64-bit binary floating point numbers; that means precision is lost for very large integers.
The proto3 JSON specification says that 64-bit integers should be encoded as strings, although numeric values can be parsed.

Fixes #1205


vchudnov-g (approved):
Thanks for catching this!
I believe Showcase itself should be JSON-encoding `int64` as strings, but LMK if you find otherwise.

jskeet:
Yup, it looks like it is. Thanks for the LGTM - merging now.

PR #1266, opened 2023-03-07
feat: add proto + logic for streaming sequence

leahecole:
Hi Noah! Per our conversation on Friday, @galz10 and I are proposing these changes to the proto - we _think_ it should be the same as `AttemptSequence` in the proto but the underlying logic will be different, and that the other methods are reusable. Let us know if we're off base!


schema/google/showcase/v1beta1/sequence.proto
```diff
@@ -60,6 +60,15 @@ service SequenceService {
     };
     option (google.api.method_signature) = "name";
   };
+
+    // Attempts a streaming sequence.
+    rpc AttemptStreamingSequence(AttemptStreamingSequenceRequest) returns (google.protobuf.Empty) {
+      option (google.api.http) = {
+        post: "/v1beta1/{name=sequences/*}"
```
noahdietz, line 67: Per CI: this collides with the `AttemptSequence` URI. So I think you'll want to make this a "custom method" by taking on a custom method verb to the end of the URI. Something like one of the following:
```suggestion
        post: "/v1beta1/{name=sequences/*}:stream"
```
```suggestion
        post: "/v1beta1/{name=sequences/*}:attemptStreamingSequence"
```

leahecole: Ack! We had just noticed right as you commented. We added one of those changes 😄

leahecole:
Moving back to draft status while things are failing - guessing it's because we're modifying the proto and not the logic, but that's just a guess

schema/google/showcase/v1beta1/sequence.proto
```diff
@@ -61,14 +61,14 @@ service SequenceService {
...
 
-    // Attempts a streaming sequence.
-    rpc AttemptStreamingSequence(AttemptStreamingSequenceRequest) returns (google.protobuf.Empty) {
-      option (google.api.http) = {
-        post: "/v1beta1/{name=sequences/*}"
-        body: "*"
-      };
-      option (google.api.method_signature) = "name";
+  // Attempts a streaming sequence.
+  rpc AttemptStreamingSequence(AttemptStreamingSequenceRequest) returns (google.protobuf.Empty) {
```
noahdietz, line 65: If this is a server streaming RPC, it needs to have the `stream` keyword in the response type ([docs](https://grpc.io/docs/what-is-grpc/core-concepts)).
```suggestion
  rpc AttemptStreamingSequence(AttemptStreamingSequenceRequest) returns (stream google.protobuf.Empty) {
```

noahdietz, line 65: Do consider if `google.protobuf.Empty` really is OK for testing resumption. For example, would it be hard to test that a stream was resumed half way through counting from `1` to `10`? Maybe not that hard because on the client side we can count how many responses were received. But do consider how resumption logic might take shape. Consider looking at other Server Stream RPCs in googleapis.

leahecole:
(sorry for all of the force pushes - not my norm. I was rebasing to try to get co-authors working and there was a bit of trial and error with syntax - finally figured it out!)

server/services/streaming_sequence_service.go
```diff
@@ -0,0 +1,178 @@
...
+			codes.NotFound,
+			"The sequence report with %q does not exist.",
+			name,
+		)
+	}
+
+	return report.(*pb.StreamingSequenceReport), nil
+}
+
+func clone_streamingSequence(s *pb.StreamingSequence) *pb.StreamingSequence {
```
shollyman, line 165: I think you can avoid all this with a single [proto.Clone](https://pkg.go.dev/google.golang.org/protobuf/proto#Clone)?

server/services/streaming_sequence_service.go
```diff
@@ -0,0 +1,178 @@
...
+	copy(r, s.GetResponses())
+
+	return &pb.StreamingSequence{
+		Name:      s.GetName(),
+		Content:   s.GetContent(),
+		Responses: r,
+	}
+}
+
+func streaming_report(n string) string {
```
noahdietz, line 176: Drive by review: `streamingReport` or `streamingReportName`

Do not use underscores in function names.

server/services/streaming_sequence_service.go
leahecole: Just for more context, I think @galz10 was looking to replicate the logic for the [unary version] of this method. We can def try replace with proto.Clone in one or both spots if appropriate - up to @noahdietz here

server/services/streaming_sequence_service.go
```diff
@@ -0,0 +1,178 @@
...
+)
+
+func (s *sequenceServerImpl) CreateStreamingSequence(ctx context.Context, in *pb.CreateStreamingSequenceRequest) (*pb.StreamingSequence, error) {
+	seq := clone_streamingSequence(in.GetStreamingsequence())
+
+	// Assign Name.
+	id := s.uid.Next()
+	seq.Name = fmt.Sprintf("streamingsequences/%d", id)
+	report := &pb.StreamingSequenceReport{
+		Name: streaming_report(seq.GetName()),
```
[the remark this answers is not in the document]
leahecole: Ooo - I see this in the unary version too, good eyes. We can try to consolidate and make sure nothing unexpected happens

server/services/streaming_sequence_service.go
leahecole: Ack, will correct in both spots

server/services/streaming_sequence_service.go
noahdietz, line 1: Can all of this go in the `sequence_service.go` file? Just keeps the server implementation in one file.

server/services/streaming_sequence_service.go
noahdietz: tbh I don't remember why I wrote a bespoke `clone` method but in looking at the implementation I think it is because `proto.Clone` wasn't doing a recursive clone for other proto messages used in the message e.g. the `repeated Response responses` field. The nature of the in-memory database that Showcase uses means we have to be very particular in how we copy b.c the client libraries writing test code against the same spots in memory might modify the data and screw something up.

server/services/streaming_sequence_service.go
leahecole: I believe so - I'll experiment with it while Gal is OOO today and tomorrow

server/services/streaming_sequence_service.go
leahecole: Given this and that it seems to be working, I'm inclined to leave it as is for now!

leahecole: (leaving the comment unresolved for @galz10 to read when he's back, Gal, resolve when you've read it!)

server/services/streaming_sequence_service.go
```diff
@@ -0,0 +1,178 @@
...
+	// Get the number of attempts, which coincides with this attempt's number.
+	n := len(rep.Attempts)
+
+	// Prepare the attempt response defined by the Sequence.
+	st := status.New(codes.OK, "Successful attempt")
+	sendStatusAtIndex := 0
+	var delay time.Duration
+	responses := seq.GetResponses()
+	content := strings.Fields(seq.GetContent())
+	if l := len(responses); l > 0 && n < l {
```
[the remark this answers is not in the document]
galz10: when `l == 0` it means there are no responses so nothing should happen.

server/services/streaming_sequence_service.go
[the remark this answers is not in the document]
galz10: The first formatting is different from the `streaming_report` would it be better to put the first formatting in a function similar to `streaming_report`.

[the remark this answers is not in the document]
leahecole: talked to Gal offline, and it seems like consolidating isn't appropriate here - they're used for different things at different times

leahecole:
ok @noahdietz - I think the logic is all in! We can remove generated files if you want, but we've found when that happens, then some tests start failing. Also, we can't seem to repro the lint failures locally so if you have any words of wisdom there we'd love em!

noahdietz:
I was able to repro the `gofmt` issues. Fixing your branch via https://github.com/googleapis/gapic-showcase/pull/1290. Merge it if you can whenever you want (I don't like pushing to other people's branches)

server/services/sequence_service.go
```diff
@@ -176,3 +183,153 @@ func clone(s *pb.Sequence) *pb.Sequence {
 		Responses: r,
 	}
 }
+
+func (s *sequenceServerImpl) CreateStreamingSequence(ctx context.Context, in *pb.CreateStreamingSequenceRequest) (*pb.StreamingSequence, error) {
+	seq := cloneStreamingSequence(in.GetStreamingsequence())
+
+	// Assign Name.
+	id := s.uid.Next()
+	seq.Name = fmt.Sprintf("streamingsequences/%d", id)
```
noahdietz, line 192:
```suggestion
	seq.Name = fmt.Sprintf("streamingSequences/%d", id)
```

noahdietz (changes requested):
All in-GitHub suggestions can be addressed as one commit locally if you prefer. Easier to convey the desired change via the "suggestion" feature than to type it out.

Also, the proto changes will require regenerating everything and probably updating server code. 

Looking pretty good though.

schema/google/showcase/v1beta1/sequence.proto
```diff
@@ -84,6 +110,67 @@ message Sequence {
...
+  // A server response to an RPC Attempt in a sequence.
+  message Response {
+    // The status to return for an individual attempt.
+    google.rpc.Status status = 1;
+
+    // The amount of time to delay sending the response.
+    google.protobuf.Duration delay = 2;
+
+    // The index that the status should be sent
+    int32 sendStatusAtIndex = 3;
```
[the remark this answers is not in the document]
galz10: Replaced with response_index, so we added this to make it easier to make the stream fail at a certain point.

schema/google/showcase/v1beta1/sequence.proto
```diff
@@ -120,17 +207,46 @@ message CreateSequenceRequest {
...
 }
 
+message AttemptStreamingSequenceRequest {
+  string name = 1 [
+    (google.api.resource_reference).type = "showcase.googleapis.com/StreamingSequence",
+    (google.api.field_behavior) = REQUIRED
+  ];
+
+  // The error that is thrown after all words are sent on the stream.
+  google.rpc.Status error = 3;
```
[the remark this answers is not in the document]
galz10: i think error might be misleading, every time you load a response it sets the "error" to the status code and at the end it of the attempt the error sent so if you had `timeout, timeout, ok` the error will be loaded with timeout the first try , timeout the seconds try and then ok on the last. Maybe we should change the name to status? @leahecole what do you think ?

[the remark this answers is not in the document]
leahecole: actually, just `status` might be confusing, given that we import the grpc status package. Wdyt about `attemptStatus`?

galz10:
@noahdietz Do you know what part of the sequence.proto generates GetStreamsequence ?

noahdietz:
> @noahdietz Do you know what part of the sequence.proto generates GetStreamsequence ?

I think the RPC name is `GetStreamingSequenceReport` with the `ing` in `Streaming`. Different from what you wrote above. Is that the issue?

galz10:
i accidentally miss typed in the comment , if you look at the failing tests its not missing the `ing` in `streaming`

> > @noahdietz Do you know what part of the sequence.proto generates GetStreamsequence ?
> 
> I think the RPC name is `GetStreamingSequenceReport` with the `ing` in `Streaming`. Different from what you wrote above. Is that the issue?

schema/google/showcase/v1beta1/sequence.proto
```diff
@@ -120,17 +209,46 @@ message CreateSequenceRequest {
   Sequence sequence = 1;
 }
 
+message CreateStreamingSequenceRequest {
+  StreamingSequence streamingSequence = 1;
```
noahdietz, line 213:
```suggestion
  StreamingSequence streaming_sequence = 1;
```

noahdietz:
Oh, ok I just suggested a fix. The getters on a generated message e.g. `CreateStreamingSequence` will take the `snake_case` field name and name the function `Get{CamelCaseOfFieldName}`. That's why we use snake_case in protos for field names.

galz10:
> CreateStreamingSequence

still failing

noahdietz:
> > Oh, ok I just suggested a fix. The getters on a generated message e.g. `CreateStreamingSequence` will take the `snake_case` field name and name the function `Get{CamelCaseOfFieldName}`. That's why we use snake_case in protos for field names.
> 
> still failing

Was everything regenerated? Any time a proto is changed, the code needs to be regenerated and the handlers need to be updated

galz10:
> > > Oh, ok I just suggested a fix. The getters on a generated message e.g. `CreateStreamingSequence` will take the `snake_case` field name and name the function `Get{CamelCaseOfFieldName}`. That's why we use snake_case in protos for field names.
> > 
> > 
> > still failing
> 
> Was everything regenerated? Any time a proto is changed, the code needs to be regenerated and the handlers need to be updated

yes it was regenerated, even if i change the `GetStreamingsequence` to `GetStreamingSequence` and regenerate it generates `GetStreamingsequence`

galz10:
@noahdietz So it looks like line 53 in sequence.proto needs to be `streaming_sequence`

noahdietz:
Update your branch as #1293 has been merged. Then reinstall `genrest` and regenerate locally.

schema/google/showcase/v1beta1/sequence.proto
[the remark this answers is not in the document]
galz10: changed it to attempt_status

noahdietz (commented):
Implementation is looking great, just the one question on the proto

schema/google/showcase/v1beta1/sequence.proto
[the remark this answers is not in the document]
galz10: I removed the status and it did not affect the code so i'll remove this.

noahdietz (approved):
Great job the both of you.

PR #1282, opened 2023-03-24
fix(rest): properly handle string-encoded well-known types in URLs

vchudnov-g:
Well-known type messages that JSON-encode to strings appear in URL paths and query params unquoted. When we extract them from these places, we need to quote them to make them valid JSON that we can pass to the JSON deserializer. We weren't doing the quoting previously.

Fixes #1263

PR #1294, opened 2023-04-12
fix: handle already quoted protojson wkt

noahdietz:
Since `protojson` encodes WKT with quotations already, quoting it again actually breaks parsing. This blocks gapic-generator-go's adoption of the latest release: https://github.com/googleapis/gapic-generator-go/actions/runs/4669784400/jobs/8268694956?pr=1276.

Only wrap with quotes if it isn't already.


vchudnov-g (approved):
Oh, this is getting called when parsing body fields as well? When working on #1282 I somehow convinced myself this wasn't getting called in that case and removed code I originally had to check for pre-existing parsing. Ooops. Thanks for fixing this. (And apologies for not seeing the post-merge comments on that PR; thanks for pinging me.)

noahdietz:
> Oh, this is getting called when parsing body fields as well? 

I don't think so! The test error I linked to was using a Standard Update, which encodes the `update_mask` in the query params.

> Thanks for fixing this. (And apologies for not seeing the post-merge comments on that PR; thanks for pinging me.)

All good! GitHub Notifications are kind of all over the place, I don't mind pinging (probably should've started with that tbh)

PR #1315, opened 2023-05-10
fix: Remove non-exist fields from method signature in sequence.proto

blakeli0:
The Java GAPIC generator fails with the latest showcase protos because the method signature for `AttemptStreamingSequence` contains non-exist fields from `AttemptStreamingSequenceRequest`

To prevent this from happening in the future, is it possible to set up `api-linter` or even similar presubmit checks as what we have in googe3?


noahdietz:
> To prevent this from happening in the future, is it possible to set up api-linter or even similar presubmit checks as what we have in googe3?

Probably a good idea...

noahdietz (approved):
Thanks for catching this.

noahdietz:
https://www.githubstatus.com

GitHub is having an incident. github actions are just failing for no reason, so they need to be rerun. This has happened on other repos.

blakeli0:
> https://www.githubstatus.com
> 
> GitHub is having an incident. github actions are just failing for no reason, so they need to be rerun. This has happened on other repos.

Thanks, I'll rerun failed tests and keep an eye on them

Issue #1338, opened 2023-06-22
IAM Policy Mixin for HttpJson matches incorrect Mux Handler


lqiu96:
Showcase test makes a call to `getIamPolicy` RPC: `GET http://localhost:7469/v1beta1/users/5da0e7e4:getIamPolicy`.

The userId should be `users/5da0e7e4`, but the error message (`A user with name users/5da0e7e4:getIamPolicy not found.`) suggests that  the userId is matched as `users/5da0e7e4:getIamPolicy`

Local Gapic Showcase Server Error:
```
2023/06/22 11:32:55 Received GET request matching '/v1beta1/{name=users/*}': "/v1beta1/users/bbe489bd:getIamPolicy"
2023/06/22 11:32:55   urlPathParams (expect 1, have 1): map["name":"users/bbe489bd:getIamPolicy"]
2023/06/22 11:32:55   request: {
  "name": "users/bbe489bd:getIamPolicy"
}
2023/06/22 11:32:55 A user with name users/bbe489bd:getIamPolicy not found.
```
It looks to have matched with this handleFunc: https://github.com/googleapis/gapic-showcase/blob/082b527b3b91c890bd3951d44e18fdfc76cdb5a3/server/genrest/genrest.go#L56 instead of https://github.com/googleapis/gapic-showcase/blob/082b527b3b91c890bd3951d44e18fdfc76cdb5a3/server/genrest/genrest.go#L105

Full Java Stacktrace:
```
com.google.api.gax.rpc.NotFoundException: Not Found

	at com.google.api.gax.rpc.ApiExceptionFactory.createException(ApiExceptionFactory.java:50)
	at com.google.api.gax.httpjson.HttpJsonApiExceptionFactory.createApiException(HttpJsonApiExceptionFactory.java:76)
	at com.google.api.gax.httpjson.HttpJsonApiExceptionFactory.create(HttpJsonApiExceptionFactory.java:54)
	at com.google.api.gax.httpjson.HttpJsonExceptionCallable$ExceptionTransformingFuture.onFailure(HttpJsonExceptionCallable.java:97)
	at com.google.api.core.ApiFutures$1.onFailure(ApiFutures.java:84)
	at com.google.common.util.concurrent.Futures$CallbackListener.run(Futures.java:1132)
	at com.google.common.util.concurrent.DirectExecutor.execute(DirectExecutor.java:31)
[...]

PR #1340, opened 2023-06-26
fix: Update HttpJson regex for URL paths

lqiu96:
Fixes: https://github.com/googleapis/gapic-showcase/issues/1338


vchudnov-g (approved):
Thanks for doing this!

noahdietz (approved):

PR #1351, opened 2023-07-07
fix: remove content_sent from StreamingSequences

galz10:
Currently, server streaming APIs do not resume from the point of failure. Before removing the contentSent property, the server would keep track of the last content that was sent before failing. To get closure to how server streams currently react to failure, we removed the contentSent property. This means that when we retry the request, the server will send the entire content, rather than starting from where it left off.

### DIFF
--- a/schema/google/showcase/v1beta1/sequence.proto
+++ b/schema/google/showcase/v1beta1/sequence.proto
@@ -166,7 +166,6 @@ message StreamingSequenceReport {
     // The status returned to the attempt.
     google.rpc.Status status = 5;
 
-    string content_sent = 6;
   }
 
   // The set of RPC attempts received by the server for a Sequence.
--- a/server/services/sequence_service.go
+++ b/server/services/sequence_service.go
@@ -50,7 +50,6 @@ type sequenceServerImpl struct {
 
 	streamingSequences sync.Map
 	streamingReports   sync.Map
-	sentContent        string
 }
 
 func (s *sequenceServerImpl) CreateSequence(ctx context.Context, in *pb.CreateSequenceRequest) (*pb.Sequence, error) {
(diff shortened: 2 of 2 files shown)


noahdietz:
Please run `gofmt -w ./server/services` to fix the lint failure

server/services/sequence_service.go
```diff
@@ -288,12 +281,8 @@ func (s *sequenceServerImpl) AttemptStreamingSequence(in *pb.AttemptStreamingSeq
 		ResponseTime:  rpb,
 		AttemptDelay:  attDelay,
 		Status:        st.Proto(),
-		ContentSent:   s.sentContent,
```
noahdietz, line 291: Is `ContentSent` part of the schema? If it is, should it be removed?

noahdietz (commented):
Can you include a little bit more detail the PR description? It's a tad vague how this get's closer to something (which is undefined in the description)

galz10:
> Can you include a little bit more detail the PR description? It's a tad vague how this get's closer to something (which is undefined in the description)

updated description

schema/google/showcase/v1beta1/sequence.proto
```diff
@@ -166,7 +166,6 @@ message StreamingSequenceReport {
     // The status returned to the attempt.
     google.rpc.Status status = 5;
 
-    string content_sent = 6;
```
noahdietz, line 169: Note: Since we discussed it offline, _this_ change will precipitate a regeneration PR - that is OK, that is intentional.

noahdietz (approved):

schema/google/showcase/v1beta1/sequence.proto
galz10: Yea thanks for clarifying!

PR #1364, opened 2023-08-10
fix: add fail index to streaming sequence

leahecole:
Gal was working on this before he moved to another team and I am opening this to make sure I tie up loose ends by finishing his work. I added a test and then I regenerated the support files because of the proto change. It seemed to regenerate more than just my service, but I removed anything unrelated to sequence. If I should re-include those, let me know. cc @galz10


noahdietz (approved):
I don't think you need to include the generated code changes in this PR. They should be automatically included in CI for testing, then contributed back via an automatic PR upon merge.

schema/google/showcase/v1beta1/sequence.proto
```diff
@@ -217,14 +217,19 @@ message AttemptSequenceRequest {
...
 
 message AttemptStreamingSequenceRequest {
   string name = 1 [
     (google.api.resource_reference).type = "showcase.googleapis.com/StreamingSequence",
     (google.api.field_behavior) = REQUIRED
   ];
-
+  
+  int32 fail_index = 2 [
+    (google.api.resource_reference).type = "showcase.googleapis.com/Sequence",
```
blakeli0, line 230: @noahdietz Is this is a valid resource reference? I think the field reference another resource should be of String type per [AIP-122](https://google.aip.dev/122), also the [Sequence](https://github.com/googleapis/gapic-showcase/blob/fe2fe253b030604605ea926b7e4664585669969d/schema/google/showcase/v1beta1/sequence.proto#L92-L95) resource is of pattern `sequences/{sequence}` which seems do not fit here.

schema/google/showcase/v1beta1/sequence.proto
```diff
@@ -217,14 +217,19 @@ message AttemptSequenceRequest {
...
 }
 
 message AttemptStreamingSequenceRequest {
   string name = 1 [
     (google.api.resource_reference).type = "showcase.googleapis.com/StreamingSequence",
     (google.api.field_behavior) = REQUIRED
   ];
-
+  
+  int32 fail_index = 2 [
```
blakeli0, line 229: Can we add some comments to explain what this field is used for? Also based on the server logic, this looks more like a `last_failed_index` or `restart_index`?

schema/google/showcase/v1beta1/sequence.proto
noahdietz: Oh yeah definitely not.

noahdietz, line 230:
```suggestion
  int32 fail_index = 2 [
```

schema/google/showcase/v1beta1/sequence.proto
leahecole: good call, I like `last_failed_index`

schema/google/showcase/v1beta1/sequence.proto
leahecole: good catch! I accepted Noah's suggestion which iiuc just removed what used to be line 30 with this reference type. I added a clarifying comment, good call @blakeli0. I want to make sure I'm interpreting these comments and the code suggestion from @noahdietz correctly though - do I not need to include a resource type at all? Or, should I be linking to something like https://github.com/googleapis/gapic-showcase/blob/3188c8a6089597d937ed6ab5b529716ab92d5ff3/schema/google/showcase/v1beta1/sequence.proto#L233?

noahdietz: Nope, not necessary. The resource here is the Sequence being attempted. The `content` is just data.

blakeli0 (approved):

PR #1385, opened 2023-11-03
feat: add EchoErrorDetails RPC returning error.details as field type google.protobuf.Any 

vchudnov-g:
In the DIREGAPIC flow (`rest`-transport clients generated from synthetic protos emitted by  https://github.com/googleapis/disco-to-proto3-converter from public Discovery documents) currently used for the Google Compute Engine GAPIC, we only allow `google.protobuf.Any` types for fields whose paths end in `.error.details`. The actual types at run-time must be one of the types in `google/rpc/error_details.proto`. Since `error_details.proto` is not explicitly included in the synthetic GCE protos or the Showcase protos, we want to ensure that GAPICs in all languages can successfully decode these `error_details.proto` messages in a language-appropriate manner.

This change implements an RPC `EchoErrorDetails()` that adds such an `Any`-type field to the schema and return such an allowed `error_details.proto` actual value from the server. This can be used to verify that generated GAPICs can indeed process `Any` fields whose actual types are those in `error_details.proto`.

The new server functionality can be tested from the command line via the following:

```
curl -i -X POST http://localhost:7469/v1beta1/echo:error-details \
   -d'{"singleDetailText": "Rain","multiDetailText":["rain", "snow" , "hail", "sleet", "fog"]}' \
   -H "Content-Type: application/json" -H "X-Goog-Api-Client: rest/0.0.0 gapic/0.0.0"     
curl -i -X POST http://localhost:7469/v1beta1/echo:error-details \
   -d'{"singleDetailText": "Rain"}' \
   -H "Content-Type: application/json" -H "X-Goog-Api-Client: rest/0.0.0 gapic/0.0.0"     
curl -i -X POST http://localhost:7469/v1beta1/echo:error-details \
   -d'{"multiDetailText":["rain", "snow" , "hail", "sleet", "fog"]}' -H "Content-Type: application/json" \
   -H "X-Goog-Api-Client: rest/0.0.0 gapic/0.0.0"     

```

(This change is motivated by https://github.com/googleapis/disco-to-proto3-converter/pull/102, which adds support for `Any` fields only in `....error.details`.)

### DIFF
--- a/schema/google/showcase/v1beta1/echo.proto
+++ b/schema/google/showcase/v1beta1/echo.proto
@@ -87,7 +87,7 @@ service Echo {
   // This method returns error details in a repeated "google.protobuf.Any"
   // field. This method showcases handling errors thus encoded, particularly
   // over REST transport. Note that GAPICs only allow the type
-  // "google.protobuf.Any" for field paths ending in "error.details", and at
+  // "google.protobuf.Any" for field paths ending in "error.details", and, at
   // run-time, the actual types for these fields must be one of the types in
   // google/rpc/error_details.proto.
   rpc EchoErrorDetails(EchoErrorDetailsRequest) returns (EchoErrorDetailsResponse) {
@@ -100,7 +100,7 @@ service Echo {
   // This method returns error details in a singular "google.protobuf.Any"
   // field. This method showcases handling errors thus encoded, particularly
   // over REST transport. Note that GAPICs only allow the type
-  // "google.protobuf.Any" for field paths ending in "error.details", and at
+  // "google.protobuf.Any" for field paths ending in "error.details", and, at
   // run-time, the actual types for these fields must be one of the types in
   // google/rpc/error_details.proto.
   rpc EchoErrorSingleDetail(EchoErrorSingleDetailRequest) returns (EchoErrorSingleDetailResponse) {
--- a/schema/google/showcase/v1beta1/echo.proto
+++ b/schema/google/showcase/v1beta1/echo.proto
@@ -237,14 +237,19 @@ message EchoResponse {
 message EchoErrorDetailsRequest {
   repeated string text = 1;
   string single_detail_text = 2;
+  repeated string multi_detail_text = 3;
 }
 // The response message used for the EchoErrorDetails method.
 message EchoErrorDetailsResponse {
   message SingleDetail {
     ErrorWithSingleDetail error = 1;
   }
+  message MultipleDetails {
+    ErrorWithMultipleDetails error = 1;    
+  }
   ErrorWithMultipleDetails error = 1;
   SingleDetail single_detail = 2;
+  MultipleDetails multiple_details = 3;
 }
 
 message ErrorWithMultipleDetails {
--- a/schema/google/showcase/v1beta1/echo.proto
+++ b/schema/google/showcase/v1beta1/echo.proto
@@ -97,19 +97,6 @@ service Echo {
     };
   }
 
-  // This method returns error details in a singular "google.protobuf.Any"
-  // field. This method showcases handling errors thus encoded, particularly
-  // over REST transport. Note that GAPICs only allow the type
-  // "google.protobuf.Any" for field paths ending in "error.details", and, at
-  // run-time, the actual types for these fields must be one of the types in
-  // google/rpc/error_details.proto.
-  rpc EchoErrorSingleDetail(EchoErrorSingleDetailRequest) returns (EchoErrorSingleDetailResponse) {
-    option (google.api.http) = {
-      post: "/v1beta1/echo:error-single-detail"
-      body: "*"
-    };
-  }
-
   // This method splits the given content into words and will pass each word back
   // through the stream. This method showcases server-side streaming RPCs.
   rpc Expand(ExpandRequest) returns (stream EchoResponse) {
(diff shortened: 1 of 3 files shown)


vchudnov-g:
Interesting. The CI failures do not _seem_ to be caused by my changes, but how else would they have sneaked in? And yet at first glance I'm not seeing the erroring source, and even the `Unchanged files with check annotations (Beta)` section of the `Files changed` pane does not show the quoted error source matching the error in the message.

noahdietz (commented):
I'm not quite sure how this is different from using the standard `Echo` request with the `error` field set. My guess is that this focuses on the `Any` being in the response body as opposed to the `Status` returned? If so, why not just put some fields on `Echo` to funnel the response through `Any`? e.g. (forgive the names) `EchoRequest.any_response_enabled` and `EchoResponse.any_content` (could be repeated).

vchudnov-g:
Yes, the big difference is that these RPCs use `Any` fields in their responses. For the purpose of the checks that motivate this PR, we're not strictly checking that they occur in RPC *error* responses, but merely that they occur in fields with paths ending in `.error.details`and that the GAPICs in various languages can respond to such fields appropriately (motivating PR: https://github.com/googleapis/disco-to-proto3-converter/pull/102).

As to changing existing RPCs to use the `Any` fields instead of adding new RPCs:
**For erroring reponses**:
For `rest` transport, at least, erroring RPCs go through [`ErrorResponse`](https://github.com/googleapis/gapic-showcase/blob/c6512469703a716c619058249f2bb55d81c5ee37/util/genrest/resttools/error_response.go#L68), which uses [`googleapi.Error`](https://pkg.go.dev/google.golang.org/api/googleapi#Error), which in turn includes an [`ErrorItem`](https://pkg.go.dev/google.golang.org/api/googleapi#ErrorItem) that does not use `Any` fields. For `grpc` transport, I admit I'm not 100% clear on how the `error` from the server gets translated into a full error type message for the client to consume. Regardless, I'm reluctant to muck around with the existing error responses without a bigger discussion, which we may well need to have soon for the general Showcase testing infrastructure.

**Extending the current `Echo` RPC**:
IIUC your suggestion correctly, you're suggesting I fold the functionality currently in this PR into the existing `Echo` RPC: adding fields in the request to signal populating a singular or repeated error details (without marking the RPC as erroring via `EchoRequest.error`), and then changing the response to include those singular or repeated error details. I could certainly do that, though I shied away from it because I did not want to cause confusion by having these error details fields (currently only used for testing `Any` field decoding) separate from the RPC error setter `EchoRequest.error` and the gRPC error response. I thought (and still feel, TBH) that having this new functionality in separate RPCs is cleaner and more comprehensible for now.

Another option is that I could combine the two RPCs currently in the PR into a single RPC, where the request has two fields to signal a single error detail or multiple, and the response changes accordingly.  That might be slightly simpler, though it would involve an additional level of nesting in the response since we want the path to both singular and repeated details to end with `.error.details`.

WDYT?

noahdietz:
> these RPCs use `Any` fields in their responses....checking...that they occur in fields with paths ending in `.error.details`and that the GAPICs in various languages can respond to such fields appropriately 

Sorry, this is throwing me. So REGAPIC is supposed to only allow `Any` when used in the `google.rpc.Status.details` field? That seems not right? I think I'm confusing wanting to test `Any` parsing and something specific to error details.

>  which in turn includes an [`ErrorItem`](https://pkg.go.dev/google.golang.org/api/googleapi#ErrorItem) that does not use `Any` fields. 

It should be using the [`Details` field](https://screenshot.googleplex.com/6PUY3ACxQeVFYSW), and from what I can see, REGAPIC [isn't reporting `Details` via the helper](https://screenshot.googleplex.com/Bss3z2Sz8uVsgi8). I understand not wanting to rock the existing implementation's boat necessarily, but I think the well-worn path here is that the `Details` field contains the `Any` messages we want.

I understand wanting to add bespoke RPCs to not overload the existing one, and we definitely split `Expand` from `PagedExpand` for similar reasons (one is streaming, the other is paginated, but they are super similar).

I'm just not sure what is *actually* being tested here. If it is just that REGAPICs can handle an `Any` for a well known type (e.g. those in google/rpc/error_details.proto), then let's refactor the schema to reflect that general "well known type" case e.g. `EchoAny` and identify the well known types we'd support parsing here (for now it is just those under google/rpc/error_details.proto, but tomorrow it could be custom registered one via some generator config). This gives an Any specific test, a place to add onto which simplifies adding new tests, and flexibility for the future. Separate RPCs could be fine as well, but we already have a practice of putting > 1 test case in a single RPC (i.e. oneof with `content` or `error` for Echo).

```

message EchoAnyRequest {
  string content = 1;
}

message EchoAnyResponse {
  // This wraps the content in an ErrorDetail
  google.protobuf.Any singular_any = 1;

  // This splits content on space and reports each word as an ErrorDetail
  repeated google.protobuf.Any repeated_any = 2;

  // Eventually say we add a custom Any type like ShowcaseError that is triggered via other config...
  // google.protobuf.Any custom_any = 3;
}
```

vchudnov-g:
As a result of an out-of-band conversation with @noahdietz , I corrected/clarified the PR description to better describe the motivation for this PR. I will also work on combining the two new RPCs into one for simplicity.

vchudnov-g:
Consolidated what were formerly two RPCs into a single one. Update PR title and description to match.

Still need to investigate the CI errors.

alexander-fenster (approved):

noahdietz (approved):

PR #1391, opened 2023-12-05
chore: add special characters to testing.proto comments

alicejli:
Adding some proto comments to both a service and RPC that contain special characters.  This is to help catch generator formatting issues.  (e.g. https://github.com/googleapis/sdk-platform-java/pull/2280)


burkedavison:
LGTM, but would like to have a showcase-go team member review.

schema/google/showcase/v1beta1/testing.proto
```diff
@@ -28,12 +28,18 @@ option ruby_package = "Google::Showcase::V1beta1";
...
+// 2) Resource: `pokemon/*/psychic/*`
 service Testing {
   // This service is meant to only run locally on the port 7469 (keypad digits
   // for "show").
   option (google.api.default_host) = "localhost:7469";
 
   // Creates a new testing session.
+  // Adding this comment with special characters for comment formatting tests:
+  // 1. (abra->kadabra->alakazam)
+  // 2) Resource: `pokemon/*/psychic/*`
```
blakeli0, line 42: Can we not mention `Resource` here as people may confuse it with the actual resource reference?

lqiu96:
QQ: Is there a reason why this is added as a showcase change instead of a unit test/ integration test inside the Java Generator? I think this could also been caught with a test to the `CommentFormatter`.

I'm not opposed since updating here could also help other languages catch the same issue.

blakeli0:
> QQ: Is there a reason why this is added as a showcase change instead of a unit test/ integration test inside the Java Generator? I think this could also been caught with a test to the `CommentFormatter`.
> 
> I'm not opposed since updating here could also help other languages catch the same issue.

I agree, I think we should test it in both unit tests and showcase tests, for the exact same reason you mentioned. Not integration tests though as it is dependent on the actual service protos which we don't control.

alicejli:
> > QQ: Is there a reason why this is added as a showcase change instead of a unit test/ integration test inside the Java Generator? I think this could also been caught with a test to the `CommentFormatter`.
> > I'm not opposed since updating here could also help other languages catch the same issue.
> 
> I agree, I think we should test it in both unit tests and showcase tests, for the exact same reason you mentioned. Not integration tests though as it is dependent on the actual service protos which we don't control.

I will look to add a test in https://github.com/googleapis/sdk-platform-java/pull/2280. In the meantime, @noahdietz / @vchudnov-g ; let us know if you have any concerns about the addition of these comments to the showcase test. Thanks!

PR #1415, opened 2024-01-11
chore: update release workflow to work on manual trigger

alicejli:
- Update the release workflow to work on manual invocation in case the automated release process fails
- Updates the release script to copy all files from `googleapis/google/api` so that any new protos added into that directory are automatically pulled in. This creates some extra baggage (e.g. including proto files that aren't strictly needed), but removes the need for manual maintenance when a new proto gets added in that is needed.

### DIFF
--- a/.github/workflows/assets.yaml
+++ b/.github/workflows/assets.yaml
@@ -4,10 +4,15 @@ on:
   release:
     types:
       - created
-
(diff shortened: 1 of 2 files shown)


noahdietz (approved):
Very nice :)

PR #1421, opened 2024-01-11
chore: Fix workflow typo and refactor to a single job

alicejli:
Follow up to https://github.com/googleapis/gapic-showcase/pull/1420

PR #1422, opened 2024-01-11
chore: fix syntax to get upload url

alicejli:
Follow up to: https://github.com/googleapis/gapic-showcase/pull/1421


noahdietz (approved):
love us some embedded JS

PR #1440, opened 2024-02-05
feat: add request id as part of EchoResponse

alicejli:
To make it easier to check what request ID is sent as part of the request, this PR adds it to EchoResponse so tests can check the response directly.

### DIFF
--- a/schema/google/showcase/v1beta1/echo.proto
+++ b/schema/google/showcase/v1beta1/echo.proto
@@ -224,6 +224,9 @@ message EchoResponse {
 
   // The severity specified in the request.
   Severity severity = 2;
+
+  // The request ID specified or autopopulated in the request.
+  string request_id = 3;
 }
 
 // The request message used for the EchoErrorDetails method.
--- a/server/services/echo_service.go
+++ b/server/services/echo_service.go
@@ -51,7 +51,7 @@ func (s *echoServerImpl) Echo(ctx context.Context, in *pb.EchoRequest) (*pb.Echo
 	}
 	echoHeaders(ctx)
 	echoTrailers(ctx)
-	return &pb.EchoResponse{Content: in.GetContent(), Severity: in.GetSeverity()}, nil
+	return &pb.EchoResponse{Content: in.GetContent(), Severity: in.GetSeverity(), RequestId: in.GetRequestId()}, nil
 }
 
 func (s *echoServerImpl) EchoErrorDetails(ctx context.Context, in *pb.EchoErrorDetailsRequest) (*pb.EchoErrorDetailsResponse, error) {


noahdietz:
Looks like the HTTP genrest handlers were generated with a request log line, but not a response log line: https://github.com/googleapis/gapic-showcase/blob/330a3a25eb8c4d0948860f1898fe7dcc05dc4ea5/server/genrest/echo.go#L83-L99

Where as the logs from the gRP Crequest come from this "observer" https://github.com/googleapis/gapic-showcase/blob/330a3a25eb8c4d0948860f1898fe7dcc05dc4ea5/cmd/gapic-showcase/loggers.go#L52-L70 which is registered here: https://github.com/googleapis/gapic-showcase/blob/330a3a25eb8c4d0948860f1898fe7dcc05dc4ea5/cmd/gapic-showcase/endpoint.go#L194-L198

Note sure why HTTP request/response logging is necessary here, feels like the observer should handle it, but that's probably your line of debugging.

"Simple fix" is to add response log to the generated rest handlers similar to how they have the `request: ` log atm. Probably not the "right" fix though.

alicejli:
> Looks like the HTTP genrest handlers were generated with a request log line, but not a response log line:
> 
> https://github.com/googleapis/gapic-showcase/blob/330a3a25eb8c4d0948860f1898fe7dcc05dc4ea5/server/genrest/echo.go#L83-L99
> 
> Note sure why HTTP request/response logging is necessary here, feels like the observer should handle it, but that's probably your line of debugging.
> 
> "Simple fix" is to add response log to the generated rest handlers similar to how they have the `request: ` log atm. Probably not the "right" fix though.

Dumb question - I had tried messing around withj the `genrest/echo.go` file, but ran into an issue where anything I added manually would get wiped when running ` go install ./util/cmd/... && go run ./util/cmd/compile_protos` (as expected since it's supposed to be autogenerated). But then I'm unable to see the result of any manual changes to `genrest/echo.go`. 

For instance, I tried modifying https://github.com/googleapis/gapic-showcase/blob/main/server/genrest/echo.go#L83 as a simple test to print out something different, but I wasn't able to get it to show up with just restarting the gapic-showcase server. 

And to your point about logging the response - agree it's not necessary although I guess parity with grpc logging would be good; right now though the REST response doesn't contain the new `request_id` field for some reason so was trying to see if I could pinpoint why

alicejli:
> Looks like the HTTP genrest handlers were generated with a request log line, but not a response log line:
> 
> https://github.com/googleapis/gapic-showcase/blob/330a3a25eb8c4d0948860f1898fe7dcc05dc4ea5/server/genrest/echo.go#L83-L99
> 
> Note sure why HTTP request/response logging is necessary here, feels like the observer should handle it, but that's probably your line of debugging.
> 
> "Simple fix" is to add response log to the generated rest handlers similar to how they have the `request: ` log atm. Probably not the "right" fix though.

Dumb question - I had tried messing around withj the `genrest/echo.go` file, but ran into an issue where anything I added manually would get wiped when running ` go install ./util/cmd/... && go run ./util/cmd/compile_protos` (as expected since it's supposed to be autogenerated). But then I'm unable to see the result of any manual changes to `genrest/echo.go`. 

For instance, I tried modifying https://github.com/googleapis/gapic-showcase/blob/main/server/genrest/echo.go#L83 as a simple test to print out something different, but I wasn't able to get it to show up with just restarting the gapic-showcase server. 

And to your point about logging the response - agree it's not necessary although I guess parity with grpc logging would be good; right now though the REST response doesn't contain the new `request_id` field for some reason so was trying to see if I could pinpoint why

alicejli:
@noahdietz thanks for your help! I was able to confirm that REST is indeed sending the correct response; for some reason when I was testing initially I hadn't rebuilt the binaries correctly. I think this PR is ready to go; I decided not to manually add in the logging to the REST handler as it'll get overwritten the next time the proto is updated anyway.

noahdietz:
>  I decided not to manually add in the logging to the REST handler as it'll get overwritten the next time the proto is updated anyway

You would've wanted to add it to the rest generator itself to prevent this.

noahdietz (approved):

Issue #1475, opened 2024-03-21
Bazel target `showcase_proto` fails with `google/api/field_info.proto: File not found.`


parthea:
When building the following bazel target at HEAD, I see `google/api/field_info.proto: File not found.` 

https://github.com/googleapis/gapic-showcase/blob/102c10a8f8f698c6edf40dbe3d7f03e3eef3ee68/schema/google/showcase/v1beta1/BUILD.bazel#L29-L46

Stack trace
```
google/api/field_info.proto: File not found.
schema/google/showcase/v1beta1/echo.proto:20:1: Import "google/api/field_info.proto" was not found or had errors.
Target //tests/integration:showcase_update failed to build
```

Issue #1476, opened 2024-03-21
Add a bazel integration test 


parthea:
An integration test should be added to ensure a successful build for the following bazel target

https://github.com/googleapis/gapic-showcase/blob/102c10a8f8f698c6edf40dbe3d7f03e3eef3ee68/schema/google/showcase/v1beta1/BUILD.bazel#L29-L46

See https://github.com/googleapis/gapic-showcase/issues/1475 where the bazel target could not be built at HEAD

noahdietz:
lmk if you want me to take this, otherwise, assign it to yourself if you are adding it already

PR #1477, opened 2024-03-21
fix: add missing dependency in bazel rule

parthea:
Fixes https://github.com/googleapis/gapic-showcase/issues/1475

Let me know if we should block this PR on https://github.com/googleapis/gapic-showcase/issues/1476


noahdietz (approved):

noahdietz:
> Let me know if we should block this PR on #1476

If you've verified locally that this works, that's fine with me.

PR #1479, opened 2024-03-21
fix(bazel): fix bazel build and add presubmit

### LINKED ISSUE #1476: Add a bazel integration test

noahdietz:
This does a number of things to fix the Bazel build:
1. adds `.bazeliskrc` so that this repo pins to bazel 6.3.0 (like googleapis)
2. adds bazel c++ flag to `.bazelrc` to fix C++ version compat issues (like googleapis)
3. updates rules_proto to an actual released version (yes its an `rc`, but that's better than a commit)
4. adds a missing `any_proto` dep to the proto_library target

This also adds a bazel build presubmit for the proto_library target to ensure any changes to the proto/WORKSPACE still work.

Fixes #1476


WORKSPACE.bazel
```diff
@@ -46,14 +46,14 @@ protobuf_deps()
...
-  name = "rules_proto",
-  sha256 = "6a0b76595cb8bb97cec085c6b45c24337d30615d636f4941225c05f017690b6f",
-  strip_prefix = "rules_proto-25c8af9d56980df7732b2f8639c5d43ef878a6f7",
-  urls = [
-    "https://mirror.bazel.build/github.com/bazelbuild/rules_proto/archive/25c8af9d56980df7732b2f8639c5d43ef878a6f7.tar.gz",
-    "https://github.com/bazelbuild/rules_proto/archive/25c8af9d56980df7732b2f8639c5d43ef878a6f7.tar.gz",
-  ],
+    name = "rules_proto",
+    sha256 = "903af49528dc37ad2adbb744b317da520f133bc1cbbecbdd2a6c546c9ead080b",
+    strip_prefix = "rules_proto-6.0.0-rc0",
```
parthea, line 51: Can we use the latest stable release instead? If not, please could you add a comment? 

[6.0.0-rc2](https://github.com/bazelbuild/rules_proto/releases/tag/6.0.0-rc2) exists. If an rc version is needed, can we use the latest rc instead?

parthea (approved):

WORKSPACE.bazel
noahdietz: The latest stable is much too old. I'll bump to rc2 (needed to add another dep for it to work).

noahdietz: Added a comment!

parthea (approved):

PR #1484, opened 2024-03-23
feat: Add google.api.api_version annotation to echo.proto

parthea:
This PR allows folks to run manual tests against AIP-4236 (See https://github.com/aip-dev/google.aip.dev/pull/1331)

`gapic-showcase run -v` will log the request headers in verbose mode as per https://github.com/googleapis/gapic-showcase/blob/main/CHANGELOG.md#v090--2020-04-22

I created https://github.com/googleapis/gapic-showcase/issues/1493 for echoing the `x-goog-api-version` header so that it can be used in automated integration tests.


parthea:
Wait for https://github.com/googleapis/gapic-showcase/pull/1485 which should resolve the error below which appears in https://github.com/googleapis/gapic-showcase/actions/runs/8402715704/job/23012403762?pr=1484
```
2024/03/23 15:19:14 google/showcase/v1beta1/echo.proto:47:10: Option "(google.api.api_version)" unknown. Ensure that your proto definition file imports the proto which defines the option.
```

```
Submodule path 'schema/googleapis': checked out '5c1f64e746af6e53713884df3eb7719f3956a66f'
```

PR https://github.com/googleapis/gapic-showcase/pull/1485 updates the git submodule from SHA [5c1f64e746af6e53713884df3eb7719f3956a66f](https://github.com/googleapis/googleapis/commit/5c1f64e746af6e53713884df3eb7719f3956a66f) to SHA [d81d0b9e6993d6ab425dff4d7c3d05fb2e59fa57](https://github.com/googleapis/googleapis/commit/d81d0b9e6993d6ab425dff4d7c3d05fb2e59fa57). See the diff [here](https://github.com/googleapis/googleapis/compare/5c1f64e746af6e53713884df3eb7719f3956a66f...d81d0b9e6993d6ab425dff4d7c3d05fb2e59fa57) which includes the `google.api.api_version` annotation.

noahdietz:
Adding the annotation is fine & necessary, but how do you expect generator integration tests to test that the generated code sends the proper api version?

There are two likely routes:
* the showcase server "echos" back the metadata/headers it was sent as trailing metadata/response headers so that the generator integration tests can capture them and assert what it sent was expected
* the server reports an error/fails the request if the API version isn't present in (one of) the expected places

I'd prefer the first one as it is not a breaking change.

noahdietz:
I think this is the proper place to implement additional HTTP header/query string translation assuming API version will be a System Parameter: https://github.com/googleapis/gapic-showcase/blob/0622c2f9ef33387fa107112e761886f8b8ba63e8/util/genrest/resttools/systemparam.go#L24-L47

noahdietz:
There might also be some code that needs updating in the `genrest` generator to ensure that system params end up being sent to the grpc handler implementation as grpc metadata.

noahdietz (changes requested):
See comments.

parthea:
> Adding the annotation is fine & necessary, but how do you expect generator integration tests to test that the generated code sends the proper api version?

I only performed manual tests using this updated proto.  Using `gapic-showcase run -v`, I can see that the header exists in the logs. I created https://github.com/googleapis/gapic-showcase/issues/1493 to follow up on adding support to echo back the `x-goog-api-version` header sent with the request to the showcase server.

```
2024/04/08 22:35:34 Received Unary Request for Method: /google.showcase.v1beta1.Echo/Echo
2024/04/08 22:35:34     Request headers:
2024/04/08 22:35:34       content-type: application/grpc
2024/04/08 22:35:34       user-agent: grpc-python/1.63.0.dev0 grpc-c/39.0.0 (linux; chttp2)
2024/04/08 22:35:34       x-goog-api-version: moose
```

parthea:
@noahdietz, Please could you take another look?

schema/google/showcase/v1beta1/echo.proto
```diff
@@ -43,6 +43,8 @@ service Echo {
   // This service is meant to only run locally on the port 7469 (keypad digits
   // for "show").
   option (google.api.default_host) = "localhost:7469";
+  // See https://github.com/aip-dev/google.aip.dev/pull/1331
+  option (google.api.api_version) = "moose";
```
noahdietz, line 47: Ok and just to make sure.....we don't want a "real" version here?

I understand from generator/client perspective this is an opaque string, but there is something to be said about having good test data...

noahdietz:
> I created #1493 to follow up on adding support to echo back the `x-goog-api-version` header sent with the request to the showcase server.

Thanks SGTM

schema/google/showcase/v1beta1/echo.proto
parthea: Fixed in a37732121833e0a9d589611bb2b60b32b85d1479

noahdietz (approved):

Issue #1493, opened 2024-04-08
Add support to echo back the `x-goog-api-version` header so that AIP-4236 can be tested


parthea:
See https://github.com/aip-dev/google.aip.dev/pull/1331 for AIP 4236 and [review feedback](https://github.com/googleapis/gapic-showcase/pull/1484#issuecomment-2043734889) from #1484

> the showcase server "echos" back the metadata/headers it was sent as trailing metadata/response headers so that the generator integration tests can capture them and assert what it sent was expected

PR #1509, opened 2024-04-26
feat(server): echo request headers in response over REST transport

vchudnov-g:
Towards #1493, but note that this will occur for all Showcase services, not just `Echo`. 

This can be verified manually on the shell via:

```
❯ curl -i -X POST -H "X-Goog-Api-Client: rest/0.0.0 gapic/0.0.0" -H 'Content-Type: application/json' http://localhost:7469/v1beta1/echo:echo -d'{"content":"g
reetings"}'                                                                                                                                                 
HTTP/1.1 200 OK
X-Showcase-Request-Accept: */*
X-Showcase-Request-Content-Length: 23
X-Showcase-Request-Content-Type: application/json
X-Showcase-Request-User-Agent: curl/8.5.0
X-Showcase-Request-X-Goog-Api-Client: rest/0.0.0 gapic/0.0.0
Date: Fri, 26 Apr 2024 21:33:24 GMT
Content-Length: 104

{
  "content":  "greetings",
  "severity":  "UNNECESSARY",
  "requestId":  "",
  "otherRequestId":  ""
}
```

and works even for error responses:

```
❯ curl -i -X POST -H "X-Goog-Api-Client: rest/0.0.0 gapic/0.0.0" -H 'Content-Type: application/json' http://localhost:7469/v1beta1/echo:echo -d'{"foo":"greet
ings"}'                                                                                                                                                     
HTTP/1.1 400 Bad Request
X-Showcase-Request-Accept: */*
X-Showcase-Request-Content-Length: 19
X-Showcase-Request-Content-Type: application/json
X-Showcase-Request-User-Agent: curl/8.5.0
X-Showcase-Request-X-Goog-Api-Client: rest/0.0.0 gapic/0.0.0
Date: Fri, 26 Apr 2024 21:47:05 GMT
Content-Length: 160

{"error":{"code":400,"message":"error reading body params '*': proto: (line 1:2): unknown field \"foo\"","details":null,"Body":"","Header":null,"Errors":null}
```

### DIFF
--- a/util/genrest/goviewcreator.go
+++ b/util/genrest/goviewcreator.go
@@ -93,6 +93,8 @@ func NewView(model *gomodel.Model) (*goview.View, error) {
 			source.P(`  backend.StdLog.Printf("  urlPathParams (expect %d, have %%d): %%q", numUrlPathParams, urlPathParams)`, len(allURLVariables))
 			source.P(`  backend.StdLog.Printf("  urlRequestHeaders:\n%%s", resttools.PrettyPrintHeaders(r, "    "))`)
 			source.P("")
+			source.P("  resttools.IncludeRequestHeadersInResponse(w, r)")
+			source.P("")
 			source.P("  if numUrlPathParams!=%d {", len(allURLVariables))
 			source.P(`    backend.Error(w, http.StatusBadRequest, "found unexpected number of URL variables: expected %d, have %%d: %%#v", numUrlPathParams, urlPathParams)`, len(allURLVariables))
 			source.P("    return")
--- a/util/genrest/resttools/headers.go
+++ b/util/genrest/resttools/headers.go
@@ -98,3 +98,16 @@ func PrettyPrintHeaders(request *http.Request, indentation string) string {
 	}
 	return strings.Join(lines, "\n")
 }
+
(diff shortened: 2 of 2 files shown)


parthea (approved):
LGTM. I tested this locally using the code in https://github.com/googleapis/gapic-generator-python/pull/1999

noahdietz:
FWIW I don't think @vchudnov-g had landed on a header prefix yet, at least the suggestions in the document he wrote are different than what is implemented here...

parthea:
At the time of merging this PR it seemed like we had settled on a prefix based on the comment thread in the document.

noahdietz:
> At the time of merging this PR it seemed like we had settled on a prefix based on the comment thread in the document.

You're right, @vchudnov-g had settled on that!

vchudnov-g:
One thought that occurs to me: maybe for the same reasons we have prefixes in the echoed http headers, we should also have them in the gRPC trailers that echo the gRPC request headers? Thoughts?

noahdietz:
> One thought that occurs to me: maybe for the same reasons we have prefixes in the echoed http headers, we should also have them in the gRPC trailers that echo the gRPC request headers? Thoughts?

Yeah it probably makes sense to do, but let's make sure we blast it out in release notes for those that already have trailer tests.

Issue #1519, opened 2024-07-01
JDK 21 - Compliance Client's method handlers not receiving the entire request body


lqiu96:
We are trying to include testing for JDK 21 and our tests are flaky for the Compliance client. We believe the change is behavior is because of some internal JDK changes updated how the HttpRequests are sent over the network. 

In JDK 17 and prior, the internal code had a [byte buffer of ~8k bytes](https://github.com/openjdk/jdk/blob/37043b05576c8b81b43ac41a8f06de0d0bbb3f5b/src/java.base/share/classes/sun/nio/cs/StreamEncoder.java#L45) but the new changes start the [capacity at 512 bytes](https://github.com/openjdk/jdk/blob/890adb6410dab4606a4f26a942aed02fb2f55387/src/java.base/share/classes/sun/nio/cs/StreamEncoder.java#L45). We think that this change made it so that it now takes multiple "writes" over the stream to send data over (i.e. it used to be able to be sent in one "write", but now takes 2+ "writes")

We believe that the golang io.Reader's Read() call does not block until all the "writes" have finished. Specifically this line: https://github.com/googleapis/gapic-showcase/blob/f0e32bcb7915e1caf80ec564597ba1fd54854055/server/genrest/compliance.go#L57 would Read and it is possible to only Read the first "write" (given that multiple "writes" could be made for a large request). We have logged the numBytes read (https://pkg.go.dev/io#Reader) and see that some messages see numBytes less than the expected value. 

#### Environment details

  - Programming language: Java
  - OS: Mac + Linux
  - Language runtime version: JDK 21 (I believe this is also an issue from JDK 19+)

#### Logs from Gapic Showcase Server
Error: `2024/07/01 17:05:20 error reading body params '*': proto: syntax error (line 1:513): invalid value`
```
2024/07/01 17:08:59   urlRequestHeaders:
    User-Agent: "Google-HTTP-Java-Client/1.44.2 (gzip)"
    X-Goog-Api-Client: "gl-java/21.0.1__Eclipse-Adoptium__Temurin-21.0.1-12 gapic/ gax/2.50.1-SNAPSHOT rest/"
    Accept: "*/*"
    Content-Length: "514"
    Accept-Encoding: "gzip"
    Content-Type: "application/json; charset=utf-8"
    Connection: "keep-alive"
2024/07/01 17:08:59 Num Bytes Read: 512
```

The Content-Length header says there are 514 bytes, but the reader only reports that 512 bytes are read.

[...]

PR #1520, opened 2024-07-02
fix: use io.ReadAll

burkedavison:
Attempting fix for #1519


burkedavison:
Updates based on description of #1519. 

Checks don't seem to be validating this part of the code: 
image

burkedavison:
Manually verify working with Java 21 showcase client integration tests.

server/genrest/compliance.go
```diff
@@ -55,8 +55,8 @@ func (backend *RESTBackend) HandleRepeatDataBody(w http.ResponseWriter, r *http.
 	// Intentional: Field values in the URL path override those set in the body.
 	var jsonReader bytes.Buffer
 	bodyReader := io.TeeReader(r.Body, &jsonReader)
-	rBytes := make([]byte, r.ContentLength)
-	if _, err := bodyReader.Read(rBytes); err != nil && err != io.EOF {
+	rBytes, err := io.ReadAll(bodyReader)
+	if err != nil && err != io.EOF {
```
codyoss, line 59: If switching to ReadAll we no longer need to check for io.EOF. just `err != nil`. From the docs: https://pkg.go.dev/io#ReadAll

" A successful call returns err == nil, not err == EOF. Because ReadAll is defined to read from src until EOF, it does not treat an EOF from Read as an error to be reported."

burkedavison: Fixed.

codyoss (approved):

vchudnov-g (commented):
LGTM

PR #1534, opened 2024-08-19
chore(cmd,server,util): replace use of github.com/golang/protobuf/proto

### LINKED ISSUE #1525: replace use of github.com/golang/protobuf/proto with google.golang.org/protobuf/proto
github.com/golang/protobuf/proto is deprecated, and we should use google.golang.org/protobuf package instead

julieqiu:
Fixes #1525

### DIFF
--- a/cmd/gapic-showcase/gapic-showcase.go
+++ b/cmd/gapic-showcase/gapic-showcase.go
@@ -9,8 +9,8 @@ import (
 	"os"
 
 	"github.com/golang/protobuf/jsonpb"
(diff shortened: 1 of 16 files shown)


codyoss:
@julieqiu are we still retaining some refs here? I would have expected the go.mod to have an update to remove the direct dep too.

julieqiu:
@codyoss - I only did github.com/golang/protobuf/proto in this CL, not the entire github.com/golang/protobuf module.

codyoss (approved):

PR #1538, opened 2024-09-23
chore(go.sum): ran go mod tidy

julieqiu:


vchudnov-g:
We have since run and submitted `mod` updates, so this should be obsolete now. Closing.

PR #1573, opened 2025-04-04
fix: return AIP/193-compliant errors over rest transport

vchudnov-g:
This makes Showcase comply with https://google.aip.dev/193#http11json-representation

```
❯ curl -H "X-Goog-Api-Client: rest/0.0.0 gapic/0.0.0" -H 'Content-Type: application/json' -i -X POST 'http://localhost:7469/v1beta1/echo:echo?alt=json;Benum-encoding=int' -d '{"error":4}'  # DEADLINE_EXCEEDED                                                                                                                                                                          
HTTP/1.1 400 Bad Request
X-Showcase-Request-Accept: */*
X-Showcase-Request-Content-Length: 11
X-Showcase-Request-Content-Type: application/json
X-Showcase-Request-User-Agent: curl/8.12.1-DEV
X-Showcase-Request-X-Goog-Api-Client: rest/0.0.0 gapic/0.0.0
Date: Fri, 04 Apr 2025 01:23:55 GMT
Content-Length: 124

{"error":{"code":400, "message":"error in query string: invalid semicolon separator in query", "status":"INVALID_ARGUMENT"}}           

❯ curl -H "X-Goog-Api-Client: rest/0.0.0 gapic/0.0.0" -H 'Content-Type: application/json' -i -X POST 'http://localhost:7469/v1beta1/echo:echo?alt=json%3Benum-encoding=int' -d '{"error":{"code":4,"message": "my error","details": [{"@type": "type.googleapis.com/google.rpc.ErrorInfo", "reason": "RESOURCE_AVAILABILITY", "domain": "compute.googleapis.com"}]}}'
HTTP/1.1 504 Gateway Timeout
X-Showcase-Request-Accept: */*
X-Showcase-Request-Content-Length: 180
X-Showcase-Request-Content-Type: application/json
X-Showcase-Request-User-Agent: curl/8.12.1-DEV
X-Showcase-Request-X-Goog-Api-Client: rest/0.0.0 gapic/0.0.0
Date: Fri, 04 Apr 2025 01:24:08 GMT
Content-Length: 209

{"error":{"code":504, "message":"my error", "status":"DEADLINE_EXCEEDED", "details":[{"@type":"type.googleapis.com/google.rpc.ErrorInfo", "reason":"RESOURCE_AVAILABILITY", "domain":"compute.googleapis.com"}]}}
```


vchudnov-g:
OK, augmented the test like I wanted.

Now I have to figure out why the presubmit is failing (open to any ideas).

util/genrest/resttools/error_response.go
```diff
@@ -62,18 +76,76 @@ func GRPCToHTTP(c codes.Code) int {
...
+			WriteShowcaseRESTImplementationError(w, "neither HTTP code or gRPC status are provided for ErrorResponse. Exactly one must be provided.")
+			return
+		}
+		httpResponseCode = GRPCToHTTP(grpcStatus)
+	} else {
+		if grpcStatus != NoCodeGRPC {
+			WriteShowcaseRESTImplementationError(w, "both HTTP code and gRPC status are provided for ErrorResponse. Exactly one must be provided.")
+			return
+		}
+		grpcStatus = HTTPToGRPC(httpResponseCode)
```
noahdietz, line 124: I personally prefer to avoid nesting and make conditions a bit more explicit for easier readability:
```suggestion
	if httpResponseCode == NoCodeHTTP && grpcStatus == NoCodeGRPC {
		WriteShowcaseRESTImplementationError(w, "neither HTTP code or gRPC status are provided for ErrorResponse. Exactly one must be provided.")
		return
	} else if httpResponseCode != NoCodeHTTP && grpcStatus != NoCodeGRPC {
		WriteShowcaseRESTImplementationError(w, "both HTTP code and gRPC status are provided for ErrorResponse. Exactly one must be provided.")
		return
	} else if grpcStatus != NoCodeGRPC {
		httpResponseCode = GRPCToHTTP(grpcStatus)
	} else if httpResponseCode != NoCodeHTTP {
		grpcStatus = HTTPToGRPC(httpResponseCode)
```

WDYT?

util/genrest/resttools/error_response.go
```diff
@@ -62,18 +76,76 @@ func GRPCToHTTP(c codes.Code) int {
...
+// ErrorResponse(). The code we're not supplying will be obtained from the one we do supply.
+const (
+	NoCodeGRPC codes.Code = 9999
+	NoCodeHTTP int        = -1
+)
+
+// ErrorResponse is a helper that formats the given response information, including the HTTP or gRPC
+// status code, a message, and any error detail types, into a RestError proto message and writes the
+// response as JSON.
+func ErrorResponse(w http.ResponseWriter, httpResponseCode int, grpcStatus codes.Code, message string, details ...interface{}) {
```
noahdietz, line 112:
```suggestion
func ErrorResponse(w http.ResponseWriter, httpResponseCode int, grpcStatus codes.Code, message string, details ...proto.Message{}) {
```

Should we save ourselves the type cast later on while packing the `Any` and just make this a `proto.Message`?

util/genrest/resttools/error_response.go
```diff
@@ -62,18 +76,76 @@ func GRPCToHTTP(c codes.Code) int {
...
-	data, _ := json.Marshal(googleAPIError)
+
+	for _, oneDetail := range details {
+		detailAsAny, err := anypb.New(oneDetail.(proto.Message))
+		if err != nil {
+			WriteShowcaseRESTImplementationError(w, fmt.Sprintf("could not interpret error detail as a protobuf Any: %+v", oneDetail))
+			return
+		}
+		jsonError.Error.Details = append(jsonError.Error.Details, detailAsAny)
+	}
```
noahdietz, line 142: Nit: Feel free to ignore this :) I think it is just as readable with super short variable identifiers, but there is probably a middle ground. This is totally a style thing ! no right/wrong
```suggestion
	for _, d := range details {
		da, err := anypb.New(d.(proto.Message))
		if err != nil {
			WriteShowcaseRESTImplementationError(w, fmt.Sprintf("could not interpret error detail as a protobuf Any: %+v", d))
			return
		}
		jsonError.Error.Details = append(jsonError.Error.Details, da)
	}
```

util/genrest/resttools/error_response.go
```diff
@@ -62,18 +76,76 @@ func GRPCToHTTP(c codes.Code) int {
...
+		gRPCCode = codeOverride
+	}
+	return gRPCCode
+}
+
+// Constants to make it obvious when we're not supplying either a gRPC or HTTP status code to
+// ErrorResponse(). The code we're not supplying will be obtained from the one we do supply.
+const (
+	NoCodeGRPC codes.Code = 9999
+	NoCodeHTTP int        = -1
```
noahdietz, line 106: Nit: Please put a comment on each of these exported constants

schema/google/showcase/v1beta1/rest_error.proto
```diff
@@ -0,0 +1,43 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+import "google/protobuf/any.proto";
+import "google/rpc/code.proto";
+import "google/rpc/status.proto";
```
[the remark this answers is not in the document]
vchudnov-g: Good catch! Removed.

util/genrest/resttools/error_response.go
vchudnov-g: Done.

util/genrest/resttools/error_response.go
vchudnov-g: That's what I tried initially but it doesn't work. Quoting from our AI:  "While a single `*errdetails.ErrorInfo` is a `proto.Message`, a slice of them (`[]*errdetails.ErrorInfo`) is not a ` []proto.Message`. You must manually create the interface slice."

IOW, while a concrete message satisfies the `proto.Message` interface, a slice of concrete messages does not match a slice of that interface. (Though I'm not sure why `[]interface{}` worked before, given this. Because it's empty?)

Ref: https://go.dev/wiki/InterfaceSlice

util/genrest/resttools/error_response.go
vchudnov-g: You're right, that's more Go-idiomatic and clearer. I modified the code to a version even simpler than what you suggest.

util/genrest/resttools/error_response.go
vchudnov-g: I get the argument for short identifiers, but I am generally opposed to them personally, unless maybe if they are only in really small, tight loops. And I really avoid one-character identifiers, because if there's any situation where one needs to do text-based (rather than semantic) substitution, they are pain.

Plus, I think they make the flow clearer (less cognitive load) when I'm reading the code to myself.

So I'll keep this,l if you don't mind.

util/genrest/resttools/error_response.go
vchudnov-g: Also relevant: `status.Status.Details()` returns `[]any`

https://pkg.go.dev/google.golang.org/grpc@v1.71.1/internal/status#Status.Details

noahdietz: Ah ok yeah that makes sense! Thanks for the in depth research there 😁

util/genrest/resttools/error_response.go
noahdietz: SGTM that's fine!

noahdietz (approved):

PR #1583, opened 2025-04-16
build: include code.proto in the release steps

vchudnov-g:
This should fix the "proto-assets" step in the "assets" action (eg for [this failing action](https://github.com/googleapis/gapic-showcase/actions/runs/14498515142/job/40672303266))


vchudnov-g:
The [action run](https://github.com/googleapis/gapic-showcase/actions/runs/14499267332/job/40674760700) I kicked off with this fix now passes the steps where it was failing before, though it's failing at the  "upload" step. That seems unrelated.

noahdietz (approved):

PR #1591, opened 2025-04-29
fix: gapic showcase proto compliance

bshaffer:
These are fixes that were required for the PHP client libraries to build. Notably they:

- remove "nested" references from `method_signature` annotations
- use standard paths for `resource.pattern`
- uses `repeated` instead of `map` for pagination

### DIFF
--- a/schema/google/showcase/v1beta1/echo.proto
+++ b/schema/google/showcase/v1beta1/echo.proto
@@ -361,9 +361,7 @@ message PagedExpandResponseList {
 
 message PagedExpandLegacyMappedResponse {
   // The words that were expanded, indexed by their initial character.
-  // (-- aip.dev/not-precedent: This is a legacy, non-standard pattern that violates
-  //     aip.dev/158. Ordinarily, this should be a `repeated` field, as in PagedExpandResponse. --)
-  map<string, PagedExpandResponseList> alphabetized = 1;
+  repeated PagedExpandResponseList alphabetized = 1;
 
   // The next page token.
   string next_page_token = 2;
--- a/schema/google/showcase/v1beta1/identity.proto
+++ b/schema/google/showcase/v1beta1/identity.proto
@@ -41,9 +41,7 @@ service Identity {
       post: "/v1beta1/users"
       body: "*"
     };
-    option (google.api.method_signature) = "user.display_name,user.email";
-    option (google.api.method_signature) =
-        "user.display_name,user.email,user.age,user.nickname,user.enable_notifications,user.height_feet";
+    option (google.api.method_signature) = "user";
   }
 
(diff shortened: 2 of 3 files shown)


schema/google/showcase/v1beta1/echo.proto
```diff
@@ -361,9 +361,7 @@ message PagedExpandResponseList {
 
 message PagedExpandLegacyMappedResponse {
   // The words that were expanded, indexed by their initial character.
-  // (-- aip.dev/not-precedent: This is a legacy, non-standard pattern that violates
```
vchudnov-g, line 364: I don't think we want to remove this. GCE uses this type of "mapped-paged" response in several RPCs, which is why this RPC was put in Showcase. How does PHP handle building the GCE GAPIC?

schema/google/showcase/v1beta1/identity.proto
```diff
@@ -41,9 +41,7 @@ service Identity {
       post: "/v1beta1/users"
       body: "*"
     };
-    option (google.api.method_signature) = "user.display_name,user.email";
```
vchudnov-g, line 44: These are legitimate method signatures. See https://google.aip.dev/client-libraries/4232#method-signatures_1 : 
* `If a field's name contains a period (.) character, this indicates a nested field.`
* `An RPC can provide this annotation more than once to specify multiple signatures. Order matters here`

schema/google/showcase/v1beta1/messaging.proto
```diff
@@ -283,10 +283,10 @@ message ListRoomsResponse {
 message Blurb {
   option (google.api.resource) = {
     type: "showcase.googleapis.com/Blurb"
-    pattern: "users/{user}/profile/blurbs/legacy/{legacy_user}~{blurb}"
-    pattern: "users/{user}/profile/blurbs/{blurb}"
+    pattern: "users/{user}/blurbs/{blurb}"
     pattern: "rooms/{room}/blurbs/{blurb}"
-    pattern: "rooms/{room}/blurbs/legacy/{legacy_room}.{blurb}"
+    pattern: "rooms/{room}/legacy_room/{legacy_room}/blurbs/{blurb}"
```
vchudnov-g, line 288: Why are these modified/added resource names needed? (I'll need to understand resource names more deeply in general.)

schema/google/showcase/v1beta1/echo.proto
bshaffer: That's interesting, I don't think we handle `map` for pagination in GCE. That may be a bug on our end.

Would you be able to point to an example in the protos (or in a doc) where this exists and how we should be handling it?

PR #1653, opened 2026-07-06
test: add REST test to TLS connection

hongalex:
This PR adds a test for REST transport alongside the existing gRPC test


cmd/gapic-showcase/tls_test.go
```diff
@@ -78,49 +79,86 @@ func TestConnectWithTLS(t *testing.T) {
...
 			Content: content,
 		},
 	}
 
-	var header metadata.MD
-	resp, err := c.Echo(ctx, req, gax.WithGRPCOptions(grpc.Header(&header)))
-	if err != nil {
-		t.Fatalf("Echo call failed: %v", err)
-	}
+	t.Run("gRPC", func(t *testing.T) {
```
coryan: Hmmm... Consider two test functions instead, and move the code to initialize the CA certificates and TLS to a helper function.

coryan (approved):
Consider refactoring. If you disagree just merge, otherwise fix and then merge.

cmd/gapic-showcase/tls_test.go
hongalex: I want to share the startup cost of spinning up the server for REST / gRPC tests. I'm happy to move the code to a helper function regardless though.

coryan: My guess is that the startup cost of two tests vs. one test could not be that bad since tests run concurrently. I would advise towards clarity vs. than saving a few seconds in execution time.

I could be completely wrong about the time it takes to initialize these though.

Ultimately this is your call. If you go for the two `t.Run()` calls then leave a comment explaining why.

hongalex: I added the helper function to reduce shared code, and added t.Parallel(), which decreases test time from around 5.5s -> 2.5s.

PR #1654, opened 2026-07-10
fix: correct go min version

shollyman:
Advertise a more correct version of minimum go.

Related: internal b/532160458


go.mod
```diff
@@ -59,6 +59,6 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 )
 
-go 1.25.8
```
coryan, line 62: SGTM, but can you comment as to why it is needed? ... Ah, I am guessing some CI system has 1.25.0 pre-installed and does not auto-update?

coryan (approved):

go.mod
shollyman: More context in b/532160458, but its a case of advertising an overly strict go version requirement.

PR #1657, opened 2026-07-15
feat: add resumable upload endpoint

bshaffer:
Adds a protobuf service (`google.showcase.v1beta1.ResumableUploadService`) and HTTP middleware (`server/resumableupload`) implementing the **Resumable Upload** Protocol.

### RPC Method

**`UploadMedia`**
- **Proto RPC**: `rpc UploadMedia(UploadMediaRequest) returns (UploadMediaResponse)`
- **HTTP Path**: `POST /resumable/upload/v1beta1/files:upload`
- **Purpose**: Canonical RPC method annotated with `google.api.http` enabling resumable upload session creation.

### Test Scenarios (`X-Goog-Test-Scenario`)

All test scenarios are controlled by passing custom HTTP headers (`X-Goog-Test-Scenario` and optional `X-Goog-Test-Scenario-Config`) when initiating the session (`X-Goog-Upload-Command: start`).

| Scenario Name | Description | Header Configuration | Expected Client Behavior |
| :--- | :--- | :--- | :--- |
| **Happy Path** | Default upload flow. Returns `X-Goog-Upload-Chunk-Granularity: 1` and completes upload. | None | Completes upload successfully on first attempt. |


server/resumableupload/resumableupload.go
```diff
@@ -0,0 +1,275 @@
...
+	return res
+}
+
+func containsCommand(commands []string, cmd string) bool {
+	for _, c := range commands {
+		if c == cmd {
+			return true
+		}
+	}
+	return false
```
coryan: I am not sure the function is worth it, since you can say:

```suggestion
    return slices.Contains(commands, cmd) 
```

server/resumableupload/resumableupload.go
```diff
@@ -0,0 +1,275 @@
...
+	path := r.URL.Path
+	if path == "" {
+		path = "/upload"
+	}
+
+	uploadURL := fmt.Sprintf("%s://%s%s?sid=%s", scheme, host, path, sid)
+	w.Header().Set("X-Goog-Upload-Status", "active")
+	w.Header().Set("X-Goog-Upload-URL", uploadURL)
+	w.Header().Set("X-Goog-Upload-Control-URL", uploadURL)
+	w.Header().Set("X-Goog-Upload-Chunk-Granularity", "1")
```
coryan: That seems like an unrealistic value, and maybe a common source of bugs in the libraries. Consider something more realistic, such as 256*1024.

server/resumableupload/resumableupload.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+	m.sessions.Store(sid, sess)
+
+	scheme := "http"
+	if r.TLS != nil {
+		scheme = "https"
+	}
+	host := r.Host
+	if host == "" {
+		host = "localhost:7469"
```
coryan: Do we have a constant we can use?

server/resumableupload/resumableupload.go
```diff
@@ -0,0 +1,275 @@
...
+
+	uploadURL := fmt.Sprintf("%s://%s%s?sid=%s", scheme, host, path, sid)
+	w.Header().Set("X-Goog-Upload-Status", "active")
+	w.Header().Set("X-Goog-Upload-URL", uploadURL)
+	w.Header().Set("X-Goog-Upload-Control-URL", uploadURL)
+	w.Header().Set("X-Goog-Upload-Chunk-Granularity", "1")
+	w.WriteHeader(http.StatusOK)
+}
+
+var startFailureMap sync.Map
```
coryan: Why sin't this part of `m *Manager` ?

server/resumableupload/resumableupload.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	id := atomic.AddInt64(&m.idGen, 1)
+	sid := fmt.Sprintf("scotty-sid-%d", id)
+
+	sess := &uploadSession{
+		ID:             sid,
+		Status:         "active",
+		Scenario:       scenario,
+		ScenarioConfig: config,
```
coryan: Methinks the `uploadSession` should also have a function that implements the scenario, not just an scenario name.

Then the handler can just lookup the function and call it...

server/resumableupload/resumableupload.go
```diff
@@ -0,0 +1,275 @@
...
+	host := r.Host
+	if host == "" {
+		host = "localhost:7469"
+	}
+	path := r.URL.Path
+	if path == "" {
+		path = "/upload"
+	}
+
+	uploadURL := fmt.Sprintf("%s://%s%s?sid=%s", scheme, host, path, sid)
```
coryan: The bots say you can use:


```suggestion
	scheme := "http"
	if r.TLS != nil {
		scheme = "https"
	}
	host := r.Host
	if host == "" {
		host = "localhost:7469"
	}
	path := r.URL.Path
	if path == "" {
		path = "/upload"
	}
	query := url.Values{}
	query.Add("sid", sid)
	newUrl := url.URL {
        Scheme: scheme,
        Host: host,
        Path: path,
        RawQuery: query.Encode().	    
	}

	uploadURL := newUrl.String()
```

server/resumableupload/resumableupload.go
```diff
@@ -0,0 +1,275 @@
...
+		FailureCount:        1,
+		ActionAfterFailures: "succeed",
+		AfterOffset:         0,
+	}
+	if cfgStr := r.Header.Get("X-Goog-Test-Scenario-Config"); cfgStr != "" {
+		_ = json.Unmarshal([]byte(cfgStr), &config)
+	}
+
+	// Use scenario+client specific failure counter key or session state
+	// For start requests before session creation, track start failures on manager if needed,
```
coryan: eeek... So I have to send the requests from one scenario from the same IP + port? What if my HTTP client closes the HTTP socket? Or creates a second socket?

In a different context I have see this done:

1. Use a separate endpoint to create "scenarios", creating an scenario returns a UUID
2. You pass the UUID in the requests to the upload endpoint
3. No need to track IP addresses, or have counters in global variables, the UUID can point to a per-scenario data structure.

coryan (commented):
More questions than requests for changes.  The scenario stuff seems brittle

server/resumableupload/resumableupload.go
coryan: Alternatively, maybe the scenarios that only work for the `start` call should ask for a client UUID in the Scenario-Config?

server/resumableupload/resumableupload.go
coryan: This comment still applies, we should use the library function to convert a URL to a string.

server/resumableupload/resumableupload.go
coryan: Still applies.

server/resumableupload/resumableupload.go
```diff
@@ -0,0 +1,199 @@
...
+	if !ok || sid == "" {
+		sendError(w, http.StatusNotFound, "Invalid or missing session ID", "")
+		return
+	}
+	sess := val.(*uploadSession)
+
+	sess.mu.Lock()
+	defer sess.mu.Unlock()
+
+	if containsCommand(commands, "query") {
```
coryan: Seems like these should be methods of the `uploadSession` type?

server/resumableupload/resumableupload.go
```diff
@@ -0,0 +1,199 @@
...
+	if offsetStr != "" {
+		var err error
+		offset, err = strconv.ParseInt(offsetStr, 10, 64)
+		if err != nil {
+			sendError(w, http.StatusBadRequest, "Invalid X-Goog-Upload-Offset header", "active")
+			return
+		}
+	}
+
+	if containsCommand(commands, "upload") {
```
coryan: Shouldn't this check `sess.Status` and reject the upload if the status was `final` or `cancelled` ?

server/resumableupload/resumableupload.go
```diff
@@ -0,0 +1,199 @@
...
+	"strings"
+	"sync"
+	"sync/atomic"
+)
+
+type uploadSession struct {
+	ID            string
+	CurrentOffset int64
+	Buffer        bytes.Buffer
+	Status        string
```
coryan: Looking at how this is used, it seems like it should be an enum, not a string.

server/resumableupload/resumableupload.go
```diff
@@ -0,0 +1,199 @@
...
+		return
+	}
+	sess := val.(*uploadSession)
+
+	sess.mu.Lock()
+	defer sess.mu.Unlock()
+
+	if containsCommand(commands, "query") {
+		w.Header().Set("X-Goog-Upload-Status", sess.Status)
+		w.Header().Set("X-Goog-Upload-Size-Received", strconv.FormatInt(sess.CurrentOffset, 10))
```
coryan: The implementation of `query` is incomplete:

- If the status is `cancelled` this should return `X-Goog-Upload-Status: cancelled`
- If the status is `final` this should return the same thing as a finalized upload does

coryan (changes requested):
Using `fmt.Sprintf()` to generate the URL string seems bad, there are specialized functions for it, let's use them.

The implementation needs some refactoring into methods, and the query implementation is incomplete.

server/resumableupload/resumableupload.go
bshaffer: done in 7faf0af

server/resumableupload/resumableupload.go
bshaffer: done in 7faf0af

server/resumableupload/resumableupload.go
bshaffer: done in 7faf0af

server/resumableupload/resumableupload.go
bshaffer: done in 7faf0af

server/resumableupload/resumableupload.go
bshaffer: done in 7faf0af

server/resumableupload/resumableupload.go
bshaffer: done in 7faf0af

server/resumableupload/resumableupload.go
bshaffer: done in 7faf0af

server/resumableupload/resumableupload.go
bshaffer: done in 7faf0af

server/resumableupload/resumableupload.go
bshaffer: removed in 7faf0af

server/resumableupload/resumableupload.go
```diff
@@ -0,0 +1,253 @@
...
+		}
+
+		m.handleRequest(w, r)
+	})
+}
+
+func (sess *uploadSession) query(w http.ResponseWriter) {
+	w.Header().Set("X-Goog-Upload-Status", string(sess.Status))
+	if sess.Status == statusCancelled {
+		w.WriteHeader(http.StatusOK)
```
coryan: I think `410 - Gone` is the correct status for these, but you need to check the documentation.

coryan: If the documentation does not give you a clear answer, please merge and open an issue to follow up.

coryan (approved):

server/resumableupload/resumableupload.go
bshaffer: Good catch! The [public documentation](https://docs.cloud.google.com/storage/docs/resumable-uploads) confirms that a `410 Gone` is returned on cancellation: 

> A session URI expires after one week but can be [cancelled](https://docs.cloud.google.com/storage/docs/performing-resumable-uploads#cancel-upload) prior to expiring. If you make a request using a session URI that is no longer valid, you receive one of the following errors:
> - A `410 Gone` status code if it's been less than a week since the upload was initiated.
> - A `404 Not Found` status code if it's been more than a week since the upload was initiated.

So cancelled returns `410` and expired returns `404`

coryan (approved):

PR #1662, opened 2026-08-12
feat: support --tls-groups flag to restrict allowed TLS groups on server

suzmue:
This change adds a `--tls-groups` flag to the GAPIC Showcase server to restrict allowed TLS 1.3 key-exchange groups. This enables client library developers to test that their clients properly negotiate specific key-exchange algorithms (such as post-quantum hybrid `X25519MLKEM768` or classical curves like `X25519`). This allows client libraries to have predictable tests that can verify which algorithm is selected.

The flag accepts a comma-separated list of standard IANA numeric group codepoints in hex (such as `0x11ec`) or decimal (such as `4588`). We explicitly support numeric IDs rather than string names so Showcase does not need to hardcode or maintain a list of algorithm names, allowing new key-exchange algorithms supported by future Go runtime versions to be used immediately without requiring Showcase updates.

Documentation and reference tables have been updated in `TLS_GUIDE.md`.

### DIFF
--- a/cmd/gapic-showcase/endpoint.go
+++ b/cmd/gapic-showcase/endpoint.go
@@ -51,6 +51,7 @@ type RuntimeConfig struct {
 	tlsCaCert    string
 	tlsCert      string
 	tlsKey       string
+	tlsGroups    string
 	enablePQC    *bool
 	autoTLS      bool
 	caCertFile   string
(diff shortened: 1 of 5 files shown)


cmd/gapic-showcase/endpoint.go
```diff
@@ -118,7 +119,18 @@ func createTLSConfig(config RuntimeConfig) *tls.Config {
...
+		if err != nil {
+			log.Fatalf("Failed to parse --tls-groups: %v", err)
+		}
+		baseConfig.CurvePreferences = groups
+		var groupNames []string
+		for _, g := range groups {
+			groupNames = append(groupNames, server.TLSGroupName(g))
+		}
+		stdLog.Printf("Restricted server TLS key exchange groups: %s", strings.Join(groupNames, ","))
+	} else if !pqcEnabled {
```
coryan, line 133: It seems like the new flag makes this one redundant. In a separate issue / PR: please talk to the potential users and figure out if we can remove the code to support `pqcEnabled`.

coryan (approved):

PR #1669, opened 2026-08-17
refactor: remove --enable-pqc flag in favor of --tls-groups

suzmue:
The --tls-groups flag provides more precise control over key-exchange algorithms and eliminates duplicate functionality. To achieve the same classical-only behavior previously provided by --enable-pqc=false, users can pass the classical curve IDs using --tls-groups=0x001d,0x0017,0x0018,0x0019 (or 29,23,24,25 in decimal).

The TLS guide documentation and integration tests have been updated accordingly.

### DIFF
--- a/TLS_GUIDE.md
+++ b/TLS_GUIDE.md
@@ -71,17 +71,6 @@ Provide the server cert, key, **and the CA cert** (which the server will use to 
 
 When running on **Go 1.24+**, the hybrid post-quantum key exchange **`X25519MLKEM768`** is enabled by default on the Showcase server.
 
-### Disabling PQC (Server-Side)
-
-If you need to verify classical fallback behavior, you can force the Showcase server to disable all Post-Quantum hybrid key exchanges and use only classical cryptography by starting the server with the `--enable-pqc=false` flag:
-
-```sh
-./gapic-showcase run \
-  --tls-cert certs/server.crt \
-  --tls-key certs/server.key \
-  --enable-pqc=false
-```
-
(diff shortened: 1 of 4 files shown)


TLS_GUIDE.md
```diff
@@ -102,6 +91,18 @@ The flag accepts a comma-separated list of standard [IANA Key Exchange Group IDs
...
+### Disabling PQC (Classical Only)
+
+To disable Post-Quantum hybrid key exchanges and restrict the Showcase server to classical cryptography only, specify the classical curves using `--tls-groups`:
+
+```sh
+# Restrict server to classical curves (X25519, P-256, P-384, P-521):
+./gapic-showcase run \
+  --tls-cert certs/server.crt \
+  --tls-key certs/server.key \
+  --tls-groups 0x001d,0x0017,0x0018,0x0019
```
coryan, line 103: Can we use symbolic names for these?

coryan (approved):
I see that using numeric values pre-exist this change, so it is not a blocker. I do wonder if we can do better.

TLS_GUIDE.md
suzmue: I decided to leave it out originally because I did not want to have to hardcode names (and make a new release when new groups are supported). I think we could iterate over the supported [CipherSuites](https://pkg.go.dev/crypto/tls#CipherSuites) to find the list instead. Will follow up.

suzmue: Nevermind CipherSuites is a distinct concept. Currently I see the options as hardcoding the list or iterating over all possible uint values...

